PR #18, opened 2023-02-11
Custom Wayland compositor

### LINKED ISSUE #7: Go fully headless
This needs to be fleshed out properly but ideally we don't need any of the followings:
 - a "real" compositor plugged to a monitor
 - an audio sink server (pulse/pipewire)
 - uinput, udev and any kernel support for it

If we implement a self-contained Wayland compositor and advertise it to applications we should be in a position of:
 - push the raw video/audio as the source to the gstreamer pipeline, encode it and push it via UDP to clients.
 - send client events (mouse/kb/pads) directly to applications without passing via udev

I need to learn more about this, here are a bunch of random links.

## Official tools
https://gitlab.freedesktop.org/wlroots/wlroots

## Reading material

https://drewdevault.com/2017/06/10/Introduction-to-Wayland.html
https://drewdevault.com/2018/02/17/Writing-a-Wayland-compositor-1.html

https://wayland-book.com/
https://wayland.app/protocols/
https://zamundaaa.github.io/wayland/2021/12/14/about-gaming-on-wayland.html

## Projects to look at

https://github.com/Hjdskes/cage
https://github.com/Plagman/gamescope

https://gitlab.freedesktop.org/mstoeckl/waypipe/
https://mstoeckl.com/notes/gsoc/blog.html

ABeltramo:
# TODO

- [x] run wayland compositor outside gstreamer in order to support pause/resume
- [x] avoid flooding of IDR requests
- [x] remove thread pool
- [x] add proper startup script to the Dockerfile
  - [x] Download additional Nvidia drivers if required
- [x] Allow changing the docker-socket path via ENV 
- [x] Change `config.toml`
  - [x] Add all current GOW apps as default apps
  - [x] Change the gstreamer pipelines to be a hierarchy of encoders (ex: `nvenc -> vaapi -> software`) so that we can automatically plug the right one at runtime
  - [x] Add ability to change the render node per app


src/wolf/wolf.cpp
```diff
@@ -246,6 +246,19 @@ auto setup_sessions_handlers(const immer::box<state::AppState> &app_state,
...
+
+          /* nvidia needs some extra paths */
+          if (session->app->h264_encoder == state::NVIDIA || session->app->hevc_encoder == state::NVIDIA) {
+            mounted_paths.push_back({get_env("NVIDIA_DRIVER_VOLUME_NAME", "nvidia-driver-vol"), "/usr/nvidia"});
+            full_devices.push_back("/dev/nvidia0");
+            full_devices.push_back("/dev/nvidia-modeset");
+            full_devices.push_back("/dev/nvidia-uvm");
+            full_devices.push_back("/dev/nvidia-uvm-tools");
+            full_devices.push_back("/dev/nvidiactl");
+          }
```
Drakulix, line 257: Just because you are encoding with an nvidia devices, you don't have to render on an nvidia device. Also you might have multiple nvidia devices (and need to map the corresponding `/dev/nvidia*` device).

Can we please not force this but make it optional? Or infer this information from the render-node for the wayland compositor instead?

ABeltramo: Both very valid points, I knew I had to move it behind a flag or something.  
The idea is to try to avoid having users to mess with the config file and provide some sensible default, in this case, since we know that nvidia is working we can safely add those.

I agree with your point, though. I'll think of a better way to (de)activate this mode!

Drakulix: I guess this plays into being able to set a custom render-node per app? So we could use the encoding-device as a default and overriding the render-node would disable automatic mapping or invoke different logic to determine the vendor of the node.

QuaiGoner (commented):
Shouldnt volume mapping for persistent user storage be:
```
~wolf/storage:/etc/wolf
~wolf/cfg/:/wolf/cfg
```
and not
`/etc/wolf/wolf:/wolf/cfg`?

PR #20, opened 2023-02-17
scroll amount is sent big-endian

zb140:
I noticed that scrolling with the mouse wheel resulted in very choppy scrolling in large chunks, so I did some digging. It looks like the scroll amount needs to be converted from big endian (compare to sunshine here: https://github.com/LizardByte/Sunshine/blob/09dff341056a26c1a41121a068327a3f8d9616df/src/input.cpp#L470).


ABeltramo (approved):
Thanks for fixing this, I've tested it out locally and it works perfectly!

PR #21, opened 2023-02-19
Input updates and horizontal scrolling

### LINKED ISSUE #16: Add horizontal scroll
There's a new horizontal [scroll input packet](https://github.com/moonlight-stream/moonlight-common-c/blob/07beb0f0a520106c49fda7664369b29e6938ea6e/src/Input.h#L113), see this pr: https://github.com/LizardByte/Sunshine/pull/793

This introduces a new `7.1.431.-1` protocol version.

zb140:
The changes in this PR are analogous to these from Sunshine, which were a _huge_ help in understanding how this stuff works 😄:
 - https://github.com/LizardByte/Sunshine/pull/511
 - https://github.com/LizardByte/Sunshine/pull/793

Previously, input packets were actually being distinguished based on packet size rather than the actual packet type fields. This worked reasonably well because nearly all packets had distinct sizes, with the notable exception of keyboard input and mouse scrolling.  This PR switches to using the correct type field instead.

In addition, there's are two new packet types: `MOUSE_HSCROLL_PACKET` for horizontal scrollwheel events, and `UTF8_TEXT_PACKET` for text events.  Horizontal scroll events are fully implemented and tested, but UTF-8 text events currently only log a warning.

Moonlight sends UTF-8 text events when it receives pasted text. On Windows, this is straightforward to implement because there is a `SendInput` API that can take Unicode text; unfortunately on Linux, there is no such API.  Instead, we'd have to "manually" convert the text to a series of key events -- I'm leaving that as an exercise for the future :grin:

### DIFF
--- a/.github/workflows/linux-build-test.yml
+++ b/.github/workflows/linux-build-test.yml
@@ -38,7 +38,7 @@ jobs:
           sudo apt-get install -y libunwind-dev
           sudo apt-get install -y \
               ninja-build \
-              libboost-thread-dev libboost-filesystem-dev libboost-log-dev libboost-stacktrace-dev \
+              libboost-locale-dev libboost-thread-dev libboost-filesystem-dev libboost-log-dev libboost-stacktrace-dev \
               libssl-dev \
               libgstreamer1.0-dev \
               libevdev-dev \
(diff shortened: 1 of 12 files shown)


zb140:
Forgot to mention: I also updated the tests and docs to match the new changes.  Also, Fixes: #16

ABeltramo (approved):
If you are happy with the changes I think we should merge this

zb140:
Sounds good to me!

Issue #9, opened 2022-12-17
Wolf logo?


ABeltramo:
Anyone wants to craft some nice logo for the project?

Murazaki:
Does it have to be a wolf ? They're quite connotated in the art world I would say, though I like FF7 Cloud's wolf design https://finalfantasy.fandom.com/wiki/Fenrir_(Final_Fantasy_VII)

![Cloud Wolf](https://static.wikia.nocookie.net/finalfantasy/images/c/cd/Cloudy_Wolf.jpg/revision/latest/scale-to-width-down/150?cb=20090501222629)

Murazaki:
![image](https://github.com/games-on-whales/wolf/assets/938538/62819fc3-0539-4474-b1d7-a53b8faf39c1)

Tried something on inkscape just now

ABeltramo:
That's cool, it looks straight out of a Monster Hunter icon
![ezgif-1-64c73bafa4](https://github.com/games-on-whales/wolf/assets/1122217/7a3ab6ce-27d5-4144-b0ba-8d2e837dbed6)
Originally the name started as the wolf that howls at the moonlight, I think it would be cool to add that to some hypothetical logo.

Murazaki:
> Originally the name started as the wolf that howls at the moonlight

I don't know why but I think it would sell great on a t-shirt 😅
Joke aside, I think we can make it as negative space on a moon background, that would work great, though not that original.

Murazaki:
![image](https://github.com/games-on-whales/wolf/assets/938538/c332281b-302c-49fe-b50c-1dd89586a932)

ABeltramo:
I quite like it!  
Would you like to open a PR and drop the file in the docs/images somewhere?

Murazaki:
No problem. It will need work to be readable at low resolution.

ABeltramo:
Done in https://github.com/games-on-whales/wolf/pull/31

PR #31, opened 2023-06-23
New wolf logo

Murazaki:
Related to issue #9


Murazaki:
oops, let me fix that bottom border first

Murazaki:
I added a
```
   style="background-color: black;"
```

To the svg document for visibility here but it might be good to either use it without it or overwrite on usage.

PR #43, opened 2023-09-10
fix: app_state_folder is empty upon pairing

### LINKED ISSUE #42: app_state_folder not using sessionID
On a fresh install of Wolf, the config file has the app_state_folder for each paired client is set to 

app_state_folder = ""

This results in that paired client using /etc/wolf as the root folder instead of /etc/woft/<sessionid>/

**Work Around**

1. Open the config.toml
2. For each paired client remove the line for app_state_folder, this will result in correct behavior

bludyh:
Small fix to set `app_state_folder` to `sessionid` upon pairing new clients. Without setting this, `app_state_folder` defaults to empty, hence the state folder does not use `sessionid` for session separation.

fix #42


ABeltramo:
Thanks for the PR, this is exactly what was needed!

PR #59, opened 2024-01-27
fix: uinput set analog trigger when zero

cancrusher:
Fix the analog trigger value never being set back to zero after being pressed.


ABeltramo:
Thanks for the PR! 

I'm also working on that, and I'm adding some better unit testing of this implementation using SDL in order to check how well the virtual devices are working. I'll better review your changes in the next few days once I've got everything properly setup.

Issue #64, opened 2024-02-17
After pulling the latest image..


AureliusGilchrist:
After pulling the latest image, I'm getting the following error:

```
[user@user-NAS:/aeternae/data_servers/wolf-GOW]$ sudo docker-compose up
[+] Running 1/1
 ✔ Container wolf-gow-wolf-1  Recreated                                                                                                                             4.6s 
Attaching to wolf-1
wolf-1  | [2024-02-17 21:33:58] 
wolf-1  | [2024-02-17 21:33:58] [ /etc/cont-init.d/10-setup_user.sh: executing... ]
wolf-1  | [2024-02-17 21:33:58] **** Configure default user ****
wolf-1  | [2024-02-17 21:33:58] Container running as root. Nothing to do.
wolf-1  | [2024-02-17 21:33:58] DONE
wolf-1  | [2024-02-17 21:33:58] 
wolf-1  | [2024-02-17 21:33:58] [ /etc/cont-init.d/15-setup_devices.sh: executing... ]
wolf-1  | [2024-02-17 21:33:58] **** Configure devices ****
wolf-1  | [2024-02-17 21:33:58] Exec device groups
wolf-1  | [2024-02-17 21:33:58] Adding user 'root' to groups: gow-gid-174
wolf-1  | [2024-02-17 21:33:59] DONE
wolf-1  | [2024-02-17 21:33:59] 
wolf-1  | [2024-02-17 21:33:59] [ /etc/cont-init.d/30-nvidia.sh: executing... ]
wolf-1  | [2024-02-17 21:33:59] 
wolf-1  | [2024-02-17 21:33:59] 
wolf-1  | [2024-02-17 21:33:59] [ /etc/cont-init.d/init-gamescope.sh: executing... ]
wolf-1  | [2024-02-17 21:33:59] **** Setting up Gamescope ****
wolf-1  | [2024-02-17 21:33:59] Launching the container's startup script as user 'root'
wolf-1  | cp: cannot create regular file '/aeternae/data_servers/wolf-GOW/sessions/fake-udev': No such file or directory
wolf-1 exited with code 0
```

What can I do to fix it? Completely vanilla docker-compose file, for the most part, too...lemme include it just in-case.                                                                    
```
version: "3.8"
services:
  wolf:
    image: ghcr.io/games-on-whales/wolf:stable
    environment:
      - XDG_RUNTIME_DIR=/tmp/sockets
      - HOST_APPS_STATE_FOLDER=./sessions
[...]

PR #65, opened 2024-03-01
doc: populate volume without error, mount /etc/wolf/cfg dir

PierreBeucher:
Hi there, I played a bit with Wolf and followed through the getting started documentation. Here's a few update suggestions for a better flow:

It's possible to create a Docker container to populate nvidia volume with `docker create` rather than use `docker run` which will exit with error - same result with a slightly better UX :)

Also Wolf expect a `/etc/wolf/cfg` directory to exists on startup and fails otherwise. This directory does not exists by default as only `/etc/wolf` is mounted. By mouting both `/etc/wolf` and `/etc/wolf/cfg` Docker will create both directories if they do not already exists, hence Wolf won't crash (as suggested here: https://github.com/games-on-whales/wolf/issues/64#issuecomment-1951479056)

### DIFF
--- a/docs/modules/user/pages/quickstart.adoc
+++ b/docs/modules/user/pages/quickstart.adoc
@@ -77,17 +77,16 @@ curl https://raw.githubusercontent.com/games-on-whales/gow/master/images/nvidia-
 
 This will create `gow/nvidia-driver:latest` locally.
 
-Unfortunately, docker doesn't seem to support directly mounting images, but you can https://docs.docker.com/storage/volumes/#populate-a-volume-using-a-container[pre-polulate volumes]!
+Unfortunately, docker doesn't seem to support directly mounting images, but you can https://docs.docker.com/storage/volumes/#populate-a-volume-using-a-container[pre-polulate volumes] by running:
 
 [source,bash]
 ....
-docker run --name nvidia-driver-container --rm --mount source=nvidia-driver-vol,destination=/usr/nvidia gow/nvidia-driver:latest sh
-
-docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown.
-ERRO[0003] error waiting for container:
+docker create --rm --mount source=nvidia-driver-vol,destination=/usr/nvidia gow/nvidia-driver:latest sh
 ....
 
-As you can see, the command will fail because the image contains no binary, but it'll create a new volume nevertheless.
+It will create a Docker container, populate `nvidia-driver-vol` with Nvidia driver if it wasn't already done and remove the container.
+
+Check volume exists with:
 
 [source,bash]
 ....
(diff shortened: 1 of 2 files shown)


docs/modules/user/pages/quickstart.adoc
```diff
@@ -133,6 +132,7 @@ docker run \
     -v nvidia-driver-vol:/usr/nvidia:rw \
     -e HOST_APPS_STATE_FOLDER=/etc/wolf \
     -v /etc/wolf:/etc/wolf:rw \
+    -v /etc/wolf/cfg:/etc/wolf/cfg:rw \
```
ABeltramo: Instead of adding an additional mount point I think it would be easier to add a line in here https://github.com/games-on-whales/wolf/blob/66e54eb5fc9afa7591d69118e16aeae0e57d405c/docker/startup.sh#L1-L8
Something like 
```bash
mkdir -p $WOLF_CFG_FOLDER
```
The docker run command is big enough already 😅

ABeltramo (commented):
Thanks for your PR, I must have missed the required folder; please see the comment

docs/modules/user/pages/quickstart.adoc
PierreBeucher: Ah indeed ! Wouldn't it be worth it to add it directly into the CPP code handling default config? Like here: https://github.com/games-on-whales/wolf/blob/66e54eb5fc9afa7591d69118e16aeae0e57d405c/src/moonlight-server/state/configTOML.cpp#L95

ABeltramo: Yeah you are right, it'd probably be better to do it in Wolf instead. Would you like to add it to your PR? It should be as simple as
```c++
std::filesystem::create_directories(utils::get_env("WOLF_CFG_FOLDER", "/etc/wolf/cfg"));
```
somewhere here https://github.com/games-on-whales/wolf/blob/66e54eb5fc9afa7591d69118e16aeae0e57d405c/src/moonlight-server/wolf.cpp#L427-L430

Otherwise, we can merge the PR and I'll add the check

PierreBeucher: I'm not familiar with cpp but I'll try to !

PierreBeucher: I tried and did not succeed in compiling CPP code sorry - there are lots of dependencies to install just to compile the project, since In run on NixOS I can't just run the commands provided in developer guide (though it's well written) and wont have time to install everything right now. 

I see you have an active Nix branch, having a `flake.nix` with development shell might help people (including myself) to contribute ! If I have time I'll try to contribute this Flake development shell :)

PierreBeucher: Added `mkdir` in container startup script as suggested in the meantime. I guess it should be easy for you to add the auto-dir create in CPP. Shall I create a small issue?

ABeltramo: Don't worry, I've got this!   
Thanks for taking the time to contribute and for this PR, looks good to me!

PR #88, opened 2024-07-25
Faking uinput: mount additional virtual devices without breaking container isolation

### LINKED ISSUE #81: Fix Steam input in unprivileged containers
Steam input works by creating virtual devices using `uinput` (similar to how we do it in [inputtino](https://github.com/games-on-whales/inputtino)). So the first step is obviously to allow the Steam container to access it:

```toml
devices = ["/dev/uinput:/dev/uinput"]
env = [
  "PROTON_LOG=1",
  "RUN_GAMESCOPE=true",
  "GOW_REQUIRED_DEVICES=/dev/uinput /dev/input/* /dev/dri/* /dev/nvidia*"
]
```

When you start a game with steam input enabled the following `uevents` will be generated:

```bash
➜ docker exec -it -u retro WolfSteam_4135727842959053255 udevadm monitor -p
monitor will print the received events for:
UDEV - the event which udev sends out after rule processing
KERNEL - the kernel uevent

KERNEL[13943.544155] add      /devices/virtual/input/input165 (input)
ACTION=add
DEVPATH=/devices/virtual/input/input165
SUBSYSTEM=input
PRODUCT=3/28de/11ff/1
NAME="Microsoft X-Box 360 pad 0"
PROP=0
EV=20000b
KEY=7cdb0000 0 0 0 0 0 0 0 0 0
ABS=3003f
FF=10000 0 0
MODALIAS=input:b0003v28DEp11FFe0001-e0,1,3,15,k130,131,133,134,136,137,13A,13B,13C,13D,13E,ra0,1,2,3,4,5,10,11,mlsf50,w
SEQNUM=23561

KERNEL[13943.544214] add      /devices/virtual/input/input165/event26 (input)
ACTION=add
DEVPATH=/devices/virtual/input/input165/event26
SUBSYSTEM=input
DEVNAME=/dev/input/event26
SEQNUM=23562
MAJOR=13
MINOR=90

KERNEL[13943.544244] add      /devices/virtual/input/input165/js3 (input)
ACTION=add
DEVPATH=/devices/virtual/input/input165/js3
SUBSYSTEM=input
DEVNAME=/dev/input/js3
SEQNUM=23563
MAJOR=13
MINOR=3
```

As you can see it's missing the corresponding `udev` events (network isolation for the win); we'll have to pick those kernel events up, `mknod` the devices and relay the correspondent `udev` event using [fake-udev](https://games-on-whales.github.io/wolf/stable/dev/fake-udev.html).  

The only bit that I'm not sure is how to avoid mounting other containers devices. Unfortunately kernel events aren't isolated, for example I can see the event for when a device is plugged in the host.. **How can we `mknod` only the devices that are actually generated by this steam instance?** Intercept `/dev/uinput` ioctl calls?

## Quick fix

This **will break isolation**, but for a single user this should work:

```toml
[[apps]]
title = "Steam"
start_virtual_compositor = true

[apps.runner]
type = "docker"
name = "WolfSteam"
image = "ghcr.io/games-on-whales/steam:edge"
mounts = ["/dev/input:/dev/input:rw", "/run/udev:/run/udev:ro"]
env = [
  "PROTON_LOG=1",
  "RUN_GAMESCOPE=true",
  "GOW_REQUIRED_DEVICES=/dev/uinput /dev/input/* /dev/dri/* /dev/nvidia*",
]
devices = ["/dev/uinput:/dev/uinput"]
ports = []
base_create_json = """
{
  "HostConfig": {
    "NetworkMode": "host",
    "IpcMode": "host",
    "CapAdd": ["SYS_ADMIN", "SYS_NICE", "SYS_PTRACE", "NET_RAW", "MKNOD", "NET_ADMIN"],
    "SecurityOpt": ["seccomp=unconfined", "apparmor=unconfined"],
    "Ulimits": [{"Name":"nofile", "Hard":10240, "Soft":10240}],
    "Privileged": false,
    "DeviceCgroupRules": ["c 13:* rmw", "c 244:* rmw"]
  }
}
\
"""
```

And set the env variable `WOLF_DOCKER_FAKE_UDEV_PATH=""` to disable our `fake_udev` implementation.

ABeltramo:
**Still a draft, will update this when I've got something that works**

Implements https://github.com/games-on-whales/wolf/issues/81

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ add_subdirectory(src/core)
 
 if (UNIX AND NOT APPLE)
     add_subdirectory(src/fake-udev)
+    add_subdirectory(src/fake-uinput)
 endif ()
 
 option(BUILD_MOONLIGHT "Build Moonlight server" ON)
--- a/src/fake-uinput/CMakeLists.txt
+++ b/src/fake-uinput/CMakeLists.txt
@@ -1,0 +1,18 @@
+option(BUILD_FAKE_UINPUT "Build fake-uinput library" ON)
+if (BUILD_FAKE_UINPUT)
+    message(STATUS "Building fake-uinput")
+
+    add_library(fake_uinput SHARED
+            fake-uinput.cpp)
+    add_library(fake-uinput::lib ALIAS fake_uinput)
+
+    target_compile_features(fake_uinput PRIVATE cxx_std_17)
+    option(FAKE_UINPUT_32BIT "Build fake-uinput 32-bit library" OFF)
+    if (FAKE_UINPUT_32BIT)
+        set_target_properties(fake_uinput PROPERTIES
+                # Force it to 32 bit
+                COMPILE_FLAGS "-m32" LINK_FLAGS "-m32"
+                # Change the name of the library to indicate it's 32-bit
+                OUTPUT_NAME "fake-uinput-32")
+    endif ()
+endif ()
--- a/src/fake-uinput/fake-uinput.cpp
+++ b/src/fake-uinput/fake-uinput.cpp
@@ -1,0 +1,98 @@
+#include <cstdarg>
+#include <cstring>
+#include <dlfcn.h>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <linux/uinput.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
+
+/** https://github.com/whot/libevdev/blob/c3953e1bb8f813f3e248047d23fb0775b147da9f/libevdev/libevdev-uinput.c#L41 **/
+#define SYS_INPUT_DIR "/sys/devices/virtual/input/"
+
+#define LOG(Thing) std::cout << "[fake-uinput] " << Thing << std::endl;
+
+typedef int (*ioctl_t)(int, unsigned long, ...);
+static ioctl_t real_ioctl = nullptr;
+
+bool load_library() {
+  LOG("Loading ...")
+  // Load the real ioctl
+  real_ioctl = reinterpret_cast<ioctl_t>(dlsym(RTLD_NEXT, "ioctl"));
+  if (!real_ioctl) {
+    LOG("Error: " << dlerror());
+    return false;
+  }
+  return true;
+}
+
+void mount(const std::filesystem::path &sysfs_path, const std::string &filename) {
+  std::filesystem::path path(filename);
+  if (!std::filesystem::exists(path)) {
+    // Get the major and minor numbers of the device
+    // They are handily available in the sysfs directory
+    std::ifstream major_minor_file(sysfs_path / "dev");
+    unsigned int major, minor;
+    char sep;
+    major_minor_file >> major >> sep >> minor;
+
+    int rc = mknod(filename.c_str(), S_IFCHR | 0666, makedev(major, minor));
+    if (rc == -1) {
+      LOG("Error creating device node: " << strerror(errno));
+    } else {
+      LOG("Created device node: " << filename);
+    }
+  }
+}
+
+/**
+ * glibc, BSD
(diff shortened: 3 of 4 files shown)


ABeltramo:
## Status

We can correctly intercept when Steam Input tries to create a virtual device, but we can't `mknod` the corresponding `/dev/input/eventX` device node from the host since we don't have enough permission when running as the default `retro` user.

To properly fix this I guess we could either:
 - Add `mknod` permissions to `retro` (how? See tests below..)
 - Add a simple CLI util, `setcap cap_mknod+ep` to it, then `system("/sudo_mknod /dev/...")`
 - Add a root running daemon inside the container that when pinged by our ld preloaded lib will do the `mknod` (+ `fake_udev` and whatever else will be needed)
 - Open a bidirectional socket to Wolf so that we can intercept that and do `docker exec -u root mknod ...`

## What I've tried so far

Build `libfake_uinput.so`
```bash
cmake --target fake_uinput
```

and `libfake-uinput-32.so`
```bash
cmake -DFAKE_UINPUT_32BIT=ON --target fake_uinput
```

Change `config.toml`:
```toml
mounts = [
  "/tmp/libfake_uinput.so:/usr/lib/x86_64-linux-gnu/fake-uinput.so:rw",
  "/tmp/libfake-uinput-32.so:/usr/lib/i386-linux-gnu/fake-uinput.so:rw",
  "/tmp/steam-startup.sh:/etc/cont-init.d/99-steam.sh:ro"
]
env = [
  "PROTON_LOG=1",
  "RUN_SWAY=true",
  "GOW_REQUIRED_DEVICES=/dev/input/* /dev/dri/* /dev/nvidia* /dev/uinput",
  "LD_PRELOAD=/usr/$LIB/fake-uinput.so"
]
devices = ["/dev/uinput:/dev/uinput"]
```

`/tmp/steam-startup.sh`:
```bash
#!/bin/bash
set -e

# Add mknod capability for Steam
setcap 'cap_mknod=eip' /usr/games/steam

# Allow user to mknod under /dev/input
mkdir -p /dev/input
chown ${PUID}:${PGID} /dev/input
```

```
cat .steam/debian-installation/logs/console-linux.txt
...
[fake-uinput] Intercepted UI_DEV_CREATE ioctl call
[fake-uinput] Error creating device node: Operation not permitted
[fake-uinput] Error creating device node: Operation not permitted
...
```

I've also tried setting `cap_mknod` to `/home/retro/.steam/debian-installation/ubuntu12_32/reaper` (and `steam` in the same subfolder) without success. Even tried with `setcap all=eip` it doesn't seem to have an effect.

zerofltexx:
Hey @ABeltramo, I had a go at finishing the shim idea from this PR and got it working on my fork.

Tried to keep it close to what you had:
- Shim only hooks UI_DEV_CREATE / UI_DEV_DESTROY, everything else passes through.
- On create it just reports the device to a new control socket endpoint (/api/v1/runtime/plug-udev-device), and Wolf does the mknod + fake-udev itself through the existing PlugDeviceEvent path, so the container doesn't need CAP_MKNOD.
- Opt-in per app with an env var (WOLF_FAKE_UINPUT=1) instead of a new runner field, so it survives the wolf-ui lobby round trip.

I didn't want to clean it into a PR before checking with you though, because the isolation part is the real problem. The container gets the actual /dev/uinput, so Steam's virtual devices end up in the host kernel. I noticed there's already work on that leak (#455), but those Wolf *virtual* seat9 rules only catch Wolf's own devices, not Steam-named ones, so my approach reopens it for the shim's devices unless they get some identity or we mediate the thing properly.

Couple questions:
1. Are you OK exposing the real /dev/uinput (opt-in, off by default), or would you rather this sit on a mediated device (CUSE, like the vuinputd stuff) that never hands out the real node?
2. If the shim's fine, how do you want its devices isolated so they don't undo the #455 fix?

Either direction works for me, just didn't want to build out a PR you'd already say no to.

PR #90, opened 2024-07-31
Prevent null pointer deference in get_mac_address

pablojrl123:
Attempting to connect over certain virtual interfaces, like Wireguard/Tailscale/VPN/etc, causes the Wolf host to throw an exception and crash.

This has been solved by ensuring the the fields of the structure returned by getifaddrs are not null.


ABeltramo:
Thanks for the PR, definitely an oversight on my side; oddly, nobody else stumbled upon it (or reported a stacktrace there).

PR #92, opened 2024-08-05
Quadlet quickstart

knoellle:
This PR adds an example quadlet file to the quickstart docs.
[Quadlets](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) are a way of describing a podman container configuration as a systemd unit.

It was adapted from what @drakulix posted on discord, updating since renamed flags/variables and bringing the state configuration in line with the other quickstart examples.

I only added this to the Intel/AMD sections since I don't have an Nvidia GPU to test with.

### DIFF
--- a/docs/modules/user/pages/quickstart.adoc
+++ b/docs/modules/user/pages/quickstart.adoc
@@ -58,6 +58,46 @@ services:
     restart: unless-stopped
 ....
 
+Podman quadlet:
+
+[source,ini]
+....
+[Unit]
+Description=Podman Wolf Gamestreaming
+
+[Service]
+TimeoutStartSec=900
+ExecStartPre=-/usr/bin/mkdir /tmp/sockets
+ExecStartPre=-/usr/bin/podman rm --force WolfPulseAudio
+Restart=on-failure
+RestartSec=5
+StartLimitBurst=5
+
+[Container]
+AutoUpdate=registry
+ContainerName=%N
+HostName=%N
+Image=ghcr.io/games-on-whales/wolf:stable
+AddCapability=CAP_SYS_PTRACE
+AddCapability=CAP_NET_ADMIN
+Network=host
+SecurityLabelDisable=true
+PodmanArgs=--ipc=host --device-cgroup-rule "c 13:* rmw"
+AddDevice=/dev/dri
+AddDevice=/dev/uinput
+AddDevice=/dev/uhid
+Environment=HOST_APPS_STATE_FOLDER=/etc/wolf
+Environment=GST_DEBUG=2
+Volume=/dev/input:/dev/input:ro
+Volume=/run/udev:/run/udev:ro
+Volume=/tmp/sockets:/tmp/sockets:rw
+Volume=/run/podman/podman.sock:/var/run/docker.sock:ro
+Volume=/etc/wolf:/etc/wolf
+
+[Install]
+WantedBy=multi-user.target
+....
+
 --
 Nvidia (Container Toolkit)::
 +


docs/modules/user/pages/quickstart.adoc
```diff
@@ -58,6 +58,46 @@ services:
     restart: unless-stopped
 ....
 
+Podman quadlet:
+
+[source,ini]
+....
+[Unit]
+Description=Podman Wolf Gamestreaming
```
robjh, line 66: This unit implies you're using a system wide podman service. This will work if you have `podman.socket` active when you run this unit, but it would be best if you added the dependency explicitly.

```
[Unit]
Description=Podman Wolf Gamestreaming
Requires=podman.service
After=podman.service 
```

docs/modules/user/pages/quickstart.adoc
```diff
@@ -58,6 +58,46 @@ services:
...
+PodmanArgs=--ipc=host --device-cgroup-rule "c 13:* rmw"
+AddDevice=/dev/dri
+AddDevice=/dev/uinput
+AddDevice=/dev/uhid
+Environment=HOST_APPS_STATE_FOLDER=/etc/wolf
+Environment=GST_DEBUG=2
+Volume=/dev/input:/dev/input:ro
+Volume=/run/udev:/run/udev:ro
+Volume=/tmp/sockets:/tmp/sockets:rw
+Volume=/run/podman/podman.sock:/var/run/docker.sock:ro
```
robjh, line 94: The `podman.socket` unit uses `%t/podman/podman.sock`.
The snippet here should do the same, since distros might put it in /var/run instead.

- https://github.com/containers/podman/blob/main/contrib/systemd/system/podman.socket
- https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Specifiers

docs/modules/user/pages/quickstart.adoc
```diff
@@ -58,6 +58,46 @@ services:
...
+AddCapability=CAP_SYS_PTRACE
+AddCapability=CAP_NET_ADMIN
+Network=host
+SecurityLabelDisable=true
+PodmanArgs=--ipc=host --device-cgroup-rule "c 13:* rmw"
+AddDevice=/dev/dri
+AddDevice=/dev/uinput
+AddDevice=/dev/uhid
+Environment=HOST_APPS_STATE_FOLDER=/etc/wolf
+Environment=GST_DEBUG=2
```
ABeltramo, line 90: Any reason why you manually specified those? It shouldn't be required since it's the same as the defaults  https://github.com/games-on-whales/wolf/blob/e10d00bba435285279fa013cad320afba97f362d/docker/wolf.Dockerfile#L115-L116

knoellle: The original quadlet file that was posted on discord set a bunch more environment variables which I removed.
Apparently I forgot to remove the `GST_DEBUG` line.
The `HOST_APPS_STATE_FOLDER` I kept because both the docker and docker-compose blocks set that variable as well.

If you want I can remove it from all of them...

docs/modules/user/pages/quickstart.adoc
```diff
@@ -58,6 +58,46 @@ services:
...
+Image=ghcr.io/games-on-whales/wolf:stable
+AddCapability=CAP_SYS_PTRACE
+AddCapability=CAP_NET_ADMIN
+Network=host
+SecurityLabelDisable=true
+PodmanArgs=--ipc=host --device-cgroup-rule "c 13:* rmw"
+AddDevice=/dev/dri
+AddDevice=/dev/uinput
+AddDevice=/dev/uhid
+Environment=HOST_APPS_STATE_FOLDER=/etc/wolf
```
detiber, line 89: Moving env var definitions to a separate file and using `EnvironmentFile` here would allow for updating env variables without having to run `systemctl daemon-reload` for the env updates to be picked up on service restart.

An example of using EnvironmentFile can be found here: https://www.flatcar.org/docs/latest/setup/systemd/environment-variables/#environmentfile-directive

I'm currently working around some permissions issues related to running on Bazzite, but here are the working configs I'm using: https://github.com/detiber/beardy-os/pull/1/files

I'm going to continue tweaking on things in that PR while I re-scope the permissions back down

ABeltramo:
Superseed by https://github.com/games-on-whales/wolf/pull/282

PR #100, opened 2024-08-25
Create a dedicated doc page for Steam

shocklateboy92:
Includes mangohud and direct steam game shortcut

I'm not sure how to run antora on this repo so I haven't been able to verify that the images work correctly yet.
Is there a github action that will build and deploy that somewhere?


ABeltramo:
Yep it's over [here](https://github.com/games-on-whales/games-on-whales.github.io) but I'm not sure this would work for external PRs.  
Thanks for taking care of this!

PR #101, opened 2024-08-26
Exposing Wolf events via API

ABeltramo:
In this PR:
- Added APIs via Unix socket to fully control Wolf programmatically [see docs](https://games-on-whales.github.io/wolf/dev-backend-api/dev/api.html)
- Added [gstinterpipe](https://github.com/RidgeRun/gst-interpipe) so that we can nicely decouple producers (running apps) from consumers (Moonlight clients) in order to implement **co-op sessions** [see docs](https://games-on-whales.github.io/wolf/dev-backend-api/user/configuration.html#_co_op_sessions) and video below
- Merged https://github.com/games-on-whales/wolf/pull/104 to add support for custom app images
- Moved to C++20, refactored events and config TOML loading to better use [reflect-cpp](https://github.com/getml/reflect-cpp)

Here's an example co-op session with two clients:

https://github.com/user-attachments/assets/25b8e78d-7661-4e4d-8f3f-be75683e8fe8


Helvio88:
What sort of sorcery is this!? 🤯

PR #104, opened 2024-08-27
feat: add support for appassets

### LINKED ISSUE #13: Display apps box arts
I've completely forgot about the `appassets` API call. We should add (optional) images as paths or base64 to the config file.

dhofer:
Fixes https://github.com/games-on-whales/wolf/issues/13 and displays box art in Moonlight.

I'm not sure where we should store these assets? Or if `asset` is even the preferred term?
Right now it's just stored relative to `HOST_APPS_STATE_FOLDER`.


ABeltramo:
Thanks for working on this it looks very good! 

The only thing I would add are some defaults images for our apps or at least one default "Wolf background" for all of them. Not sure where we could source them though..

ABeltramo:
Thank you so much for working on this, it has been merged via https://github.com/games-on-whales/wolf/pull/101

PR #108, opened 2024-08-30
Update quickstart.adoc to inform that nvidia container toolkit still needs `modeset=1`

salty2011:
Nvidia container toolkit still needs to have the nvidia driver on the host set to modeset=1 like the manual option. Just copied this across to the container toolkit


shocklateboy92 (approved):
LGTM. I was wondering if we should link to where the other steps exist instead of duplicating.

But it looks like they exist in kind of an out-of-the-way place, so maybe we're fine duplicating.

PR #111, opened 2024-09-02
Docs: Create Apps Section

salty2011:
Under the 'User Guide' section have done following

- Created a new subsection called Apps
- Moved Steam to this section
- Created a new page for pegasus so that config related doco from https://github.com/games-on-whales/gow/pull/164 is now part of this documentation.
- Updated the nav to have this new section appear under the User Guide


ABeltramo:
Thanks for working on this, looks good!

PR #125, opened 2024-09-15
Doc: keyboard layout environment variable and compositor config Sway / Gamescope

PierreBeucher:
Added a bit of doc:

- How to switch between Sway and Gamescope window compositor (I had issues with some games using Sway so I had to switch back to Gamescope, taking the opportunity to update doc)
- How to use environment variables to configure keyboard layout - works with both Gamescope and Sway whereas config file only work for Sway


ABeltramo:
Thanks for contributing, this looks good to me!

PR #179, opened 2025-03-02
Wolf UI

### LINKED ISSUE #80: A UI for Wolf
This has been discussed already many times over a few places (https://github.com/games-on-whales/wolf/issues/69#issuecomment-2094133261 https://github.com/games-on-whales/wolf/issues/44) and it's getting more and more clear that we need a UI for Wolf. The main goals are:

- **Configuration**: We need an easy way to configure Wolf without editing the config file.
- **Sharing**: Allow users to share a game session with multiple users (instead of making completely isolated sessions
  as we are doing right now).
- **Monitoring**: Allow users to see what's the status of the game session (ex: "Downloading docker image", "Container
  died", etc.)
- **Usability**: This GUI has to be controllable by keyboard and mouse, whilst at the same time should be usable in a TV
  screen even just with a gamepad.

## Implementation

I think our "dream solution" here would have:

- Ability to run multithreaded directly inside Wolf so that each user will have its own UI state, and we can show what's
  happening even before the app container is started.
- A way to easily communicate bidirectionally with Wolf (status reports, configuration changes, etc.)
    - An easy way to inject mouse, keyboard and joypad events without going through the system
- Expose direct access to the raw framebuffer where the UI is rendered so that we can feed it directly into the video
  encoding pipeline.
- _(optionally)_ An easy way to build the UI, possibly noob friendly like HTML+CSS+JS.

I've been looking around for something like this and I think the best candidate is [Sciter](https://sciter.com/), more specifically the windowless version of Sciter: [Sciter.Lite](https://sciter.com/sciter-lite-is-published/)  
There are some examples:

- [render to OpenGL](https://gitlab.com/sciter-engine/sciter-js-sdk/-/tree/main/demos.lite/sciter-glfw-opengl?ref_type=heads)
- [render to Bitmap](https://gitlab.com/sciter-engine/sciter-js-sdk/-/tree/main/demos.lite/win32-bitmap?ref_type=heads) (uses CPU rasterization)

The nice thing is that it seems to tick all the boxes above, the only concern is that it's not open source, and it's not free (albeit just for commercial use which shouldn't concern us). Is there any other alternative that we should consider?

___

Otherwise, we could just run any normal GUI toolkit (GTK, Qt, Electron, etc.) in a separate process inside the Wolf container, hook it up with our Wayland comp, and communicate with it via a socket or something. This would be a lot more complex and
less efficient, but it would be open source and free.

ABeltramo:
Plugging together https://github.com/games-on-whales/wolf-ui with Wolf.
WIP will update the PR once more things are ready


ABeltramo:
# Introducing profiles

After a lot of discussion in the dev channels, I've added a new concept in Wolf: `profiles`. This will be particularly useful for [wolf-ui](https://github.com/games-on-whales/wolf-ui) which will be the new entrypoint when starting a Moonlight streaming session and will allow to:
 - easily pick up a profile and start an app (this will allow you to carry over savegames and state between different Moonlight clients)
 - easily join another streaming session for co-op playing.

The format in `config.toml` is the following

```toml
[[profiles]]
id = 'moonlight-profile-id'

    [[profiles.apps]]
    title = 'Wolf UI'

        [profiles.apps.runner]
        run_cmd = '/home/ale/repos/wolf-ui/src/wolf-ui'
        type = 'process'

    [[profiles.apps]]
    start_audio_server = false
    start_virtual_compositor = false
    title = 'Test ball'

        [profiles.apps.audio]
        source = 'audiotestsrc wave=ticks is-live=true'

        [profiles.apps.runner]
        run_cmd = "sh -c \"while :; do echo 'running...'; sleep 10; done\""
        type = 'process'

        [profiles.apps.video]
        source = '''videotestsrc pattern=ball flip=true is-live=true !
                           video/x-raw, framerate={fps}/1''' 

[[profiles]]
id = 'user'
name = 'User'
apps = []
```

We'll automatically migrate from the current `config.toml` by:
 - Adding the hardcoded `moonlight-profile-id` which will represent the apps that are going to be shown in the Moonlight UI
 - Adding a default `user` profile and copying over all the previously defined `apps` in here.

## APIs

The currently available `/api/v1/apps` will now control the apps that will be shown in the Moonlight UI. There are 3 new endpoints to control profiles:
 - `GET /api/v1/profiles` returns a list of all defined profiles (most notably it wouldn't return the special "moonlight" profile, you'll have to use `/api/v1/apps` to interact with that one)
 - `POST /api/v1/profiles/add` takes a full profile object
 - `POST /api/v1/profiles/remove` takes just an `id`: the ID of the profile to be deleted

Hitting any of the `apps` or `profiles` endpoints will also automatically save the changes into the `config.toml` (if any).

I've yet to define how to associate `streaming` sessions with `profiles` and how that will play with CO-OP sessions, I'll start working on that now.

@Skerga, @salty2011 please let me know what you guys think and what else you would like to see added!

ABeltramo:
# Introducing Lobbies

On top of `profiles` we are also introducing Lobbies, here are the highlights:

## Key Features

- **Easy Co-op Session Joining**:
  - Users can now seamlessly join cooperative game sessions from different Moonlight clients

- **Session Continuity Across Devices**:
  - Start a gaming session on one device and continue it on another one without losing progress.

- **Watch Parties**:
  - This will also make possible "watch parties" where users can observe game sessions without interacting. This is perfect for spectators who want to follow along without affecting the gameplay.

Skerga:
Here is my list of things that would still be required for me to deem this branch beta test ready.

- [x] A rebase of the branch.
- [x] Entry in the config defining how an Apps save folder should be named. Possibly something like `directory_name` that 
       can be optional and would default to either the `title` or the `runner.name`.
       this includes maybe defining how wolf-ui should structure profile data. the currently implemented way is: 
       `(/etc/wolf/)profile-data/{profile_id}/{app.runner.name}`
- [x] The same as above but for co-op lobbies, in case a user would like to keep a separate install for co-op play,
       would default to the one above if not used.

Optional stuff that would be nice to have but can wait if necessary:

- [ ] A Keybind that would allow to return back to Wolf-Ui without directly stopping the Lobby the app is running in, 
       including leaving the lobby. (Wolf-UI has logic that allows a user to rejoin a lobby, so switching devices would be
       easier.)
- [x] Setting a Pin for a Profile. currently all profiles are open by default.
- [x] Accessing Images defined in `icon_png_path` via the wolf api if its a local png. this way i could drop access to the folder
       containing the images from wolf-ui's requirements. wolf-ui is capable of using urls by itself and does already access
       the icons on github if `icon_png_path` is not defined.
- [x] API call to pull and update docker images, with some form of progress feedback. This would allow me to keep the
       current update function without the need to access the docker socket and allow wolf-ui to be run less privileged.
- [x] an API endpoint to check if an Apps Docker Image exists on Disk, see reason above.

ABeltramo:
# Current status

Everything is pretty much wired together and working, see the short video below (youtube link):

[Video showcasing co-op lobbies](http://www.youtube.com/watch?v=8an-SvnD4pk)

There are two main issues remaining:
 - ~~On **AMD**: (I suppose Mesa in general) leaving a session on one resolution and joining from a device that asks for a bigger resolution causes a kernel panic. Not sure if this is specific to my dev machine, kernel version, etc.~~ **should be fixed now**
 - ~~On **Nvidia**: when any user stops a stream (either by leaving a lobby or by force closing from Moonlight) every other Wayland session running will crash.~~ fixed in https://github.com/games-on-whales/gst-wayland-display/pull/20

Marshy-Madness:
> # Current status
> 
> Everything is pretty much wired together and working, see the short video below (youtube link):
> 
> [Video showcasing co-op lobbies](http://www.youtube.com/watch?v=8an-SvnD4pk)
> 
> There are two main issues remaining:
> 
>     * ~On **AMD**: (I suppose Mesa in general) leaving a session on one resolution and joining from a device that asks for a bigger resolution causes a kernel panic. Not sure if this is specific to my dev machine, kernel version, etc.~ **should be fixed now**
> 
>     * On **Nvidia**: when any user stops a stream (either by leaving a lobby or by force closing from Moonlight) every other Wayland session running will crash.
>       
>       * This is tied to the fact that the zero copy video encoding pipeline needs to use some OpenGL elements (`glupload` and `glcolorconvert` to pass the DMA buffer into `nv*enc`) and when they get destroyed they'll trash some global context (I guess) that will mess up with our custom compositor (which uses OpenGL to render) causing it to crash.
>       * this issue is also present on `:stable`
>       * not sure how to get around this, I'll try to [share the OpenGL context](https://gstreamer.freedesktop.org/documentation/additional/design/opengl.html?gi-language=c) from `gst-wayland-comp` to the downstream `gl` elements, but I'm afraid it's [fairly complicated](https://discourse.gstreamer.org/t/shared-context-glmemory-and-glcolorconvert-filter-issue/1240/3)

Has this been fixed?

ABeltramo:
> Has this been fixed?

Nope, otherwise this would've been merged. :wink:

ABeltramo:
Just pushed a potential fix for the Nvidia issue, lots of changes.. Brings in github.com/games-on-whales/gst-wayland-display/pull/20 and the forked gst-interpipe at https://github.com/games-on-whales/gst-interpipe

TODO:
- [ ] we should be in full control over which Nvidia GPU gets used in a multi-GPU setup. Right now it's hardcoded to the device `0`, we should match the cuda device ID with the selected `render_node` instead.
- [ ] Test this properly! I gave it a light run with a couple sessions and lobbies, and it seems to work on my dev machine; let's see if it works when running in Docker too..

Greaper88:
For the multi-GPU setup, if say, you have 3 GPUs (my ultimate goal), would it be possible to auto-balance the load of the containers across the GPUs? IDK if that's possible to do dynamically, or if maybe it'd be possible to call the nvidia-smi command during launch to find the GPU with the lowest load and then assign that to the new container.

ABeltramo:
@Greaper88 yes, that is possible. We have full control over which GPU gets used for both the Wayland compositor and the video encoding pipeline so we could (in theory) automatically assign GPUs to different clients. It's definitely not implemented though..

ABeltramo:
Thanks everyone for following this amazing journey, it's time to merge this and start polishing and fixing the resulting issues. Time to start using `wolf:stable` everyone, see you on the other side!

PR #215, opened 2025-06-21
Add IPv6 Support for Linux

### LINKED ISSUE #136: Support on ipv6 streaming
I want to streaming over ipv6 network. However wolf only listen on ipv4 for now. Sunshine does supports IPv6, so maybe its implementation can be referred.

```
sudo ss -nlp |grep wolf
u_str LISTEN 0      4096      /tmp/sockets/wayland-1 13827879        * 0    users:(("wolf",pid=43843,fd=93)) 
u_str LISTEN 0      4096      /tmp/sockets/wayland-2 14164084        * 0    users:(("wolf",pid=43843,fd=122))
u_seq LISTEN 0      128    @cuda-uvmfd-4026533524-1@ 7927072         * 0    users:(("wolf",pid=43843,fd=10)) 
udp   UNCONN 0      0                        0.0.0.0:47999     0.0.0.0:*    users:(("wolf",pid=43843,fd=21)) 
udp   UNCONN 213760 0                        0.0.0.0:48100     0.0.0.0:*    users:(("wolf",pid=43843,fd=137))
udp   UNCONN 213760 0                        0.0.0.0:48200     0.0.0.0:*    users:(("wolf",pid=43843,fd=175))
udp   UNCONN 0      0                              *:48100           *:*    users:(("wolf",pid=43843,fd=138))
udp   UNCONN 0      0                              *:48200           *:*    users:(("wolf",pid=43843,fd=176))
tcp   LISTEN 0      4096                     0.0.0.0:47989     0.0.0.0:*    users:(("wolf",pid=43843,fd=16)) 
tcp   LISTEN 0      4096                     0.0.0.0:47984     0.0.0.0:*    users:(("wolf",pid=43843,fd=31)) 
tcp   LISTEN 0      4096                     0.0.0.0:48010     0.0.0.0:*    users:(("wolf",pid=43843,fd=20)) 
```

haohaolee:
Implements https://github.com/games-on-whales/wolf/issues/136

Changes:
- use "::" instead of "0.0.0.0" when possible
- parse ipv6 string for some formatting cases
- enhance `get_mac_address` to cover ipv6 cases

### DIFF
--- a/.github/workflows/linux-build-test.yml
+++ b/.github/workflows/linux-build-test.yml
@@ -73,6 +73,7 @@ jobs:
                 -DTEST_NVIDIA=OFF \
                 -DTEST_EXCEPTIONS=OFF \
                 -DTEST_UHID=OFF \
+                -DTEST_LINUX_NETWORK=ON \
                 -G Ninja
 
       - name: Build tests + lib
--- a/src/moonlight-server/control/control.cpp
+++ b/src/moonlight-server/control/control.cpp
@@ -41,7 +41,9 @@ enet_host create_host(std::string_view host, std::uint16_t port, std::size_t pee
   enet_address_set_host(&addr, host.data());
   enet_address_set_port(&addr, port);
 
-  auto enet_host = enet_host_create(AF_INET, &addr, peers, 0, 0, 0);
+  // host may be ipv6 or ipv4
+  // when ipv6, enet will try to enable dual-stack mode
+  auto enet_host = enet_host_create(addr.address.ss_family, &addr, peers, 0, 0, 0);
   if (enet_host == nullptr) {
     logs::log(logs::error, "An error occurred while trying to create an ENet server host.");
   }
--- a/src/moonlight-server/control/control.hpp
+++ b/src/moonlight-server/control/control.hpp
@@ -19,7 +19,7 @@ void run_control(int port,
                  const std::shared_ptr<events::EventBusType> &event_bus,
                  int peers = 20,
                  std::chrono::milliseconds timeout = 1000ms,
-                 const std::string &host_ip = "0.0.0.0");
+                 const std::string &host_ip = "::");
 
 using enet_clients_map = immer::map<ENetPeer *, immer::box<events::StreamSession>>;
 
--- a/src/moonlight-server/platforms/hw_linux.cpp
+++ b/src/moonlight-server/platforms/hw_linux.cpp
@@ -1,5 +1,6 @@
 #include "hw.hpp"
 #include <arpa/inet.h>
+#include <boost/asio/ip/address.hpp>
 #include <fcntl.h>
 #include <filesystem>
 #include <fmt/core.h>
@@ -209,6 +210,19 @@ std::string get_ip_address(ifaddrs *ifa) {
 }
 
 std::string get_mac_address(std::string_view local_ip) {
+
+  boost::asio::ip::address addr;
+  try {
+    addr = boost::asio::ip::make_address(local_ip);
+    // handle IPv6 addresses like ::ffff:127.0.0.1
+    if (addr.is_v6() && addr.to_v6().is_v4_mapped()) {
+      addr = addr.to_v6().to_v4();
+    }
+  } catch (const boost::system::system_error &e) {
+    logs::log(logs::warning, "Invalid IP address '{}': {}", local_ip, e.what());
+    return "00:00:00:00:00:00";
+  }
+
   ifaddrs *ifaddrptr = nullptr;
   if (getifaddrs(&ifaddrptr) == -1) {
     logs::log(logs::warning,
@@ -222,8 +236,23 @@ std::string get_mac_address(std::string_view local_ip) {
   // First: search for the interface name that has the same IP address
   std::string interface = "";
   for (auto ifa = ifAddrStruct.get(); ifa != NULL; ifa = ifa->ifa_next) {
-    if (ifa->ifa_addr && local_ip == get_ip_address(ifa)) {
-      interface = ifa->ifa_name;
+    if (!ifa->ifa_addr)
+      continue;
+
+    if (addr.is_v4() && ifa->ifa_addr->sa_family == AF_INET) {
+      auto &v4_addr = ((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+      auto addr_bytes = addr.to_v4().to_bytes();
+      if (memcmp(&v4_addr, addr_bytes.data(), sizeof(v4_addr)) == 0) {
+        interface = ifa->ifa_name;
+        break;
+      }
+    } else if (addr.is_v6() && ifa->ifa_addr->sa_family == AF_INET6) {
+      auto &v6_addr = ((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_addr;
+      auto addr_bytes = addr.to_v6().to_bytes();
+      if (memcmp(&v6_addr, addr_bytes.data(), sizeof(v6_addr)) == 0) {
+        interface = ifa->ifa_name;
+        break;
+      }
     }
   }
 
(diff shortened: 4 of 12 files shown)


ABeltramo:
Thanks for contributing to Wolf!  
I'm in the middle of another PR, so it might take me a few days to properly test this out. I saw you've added some unit tests for the basic methods added, brilliant job!
  
I was wondering how I can test this out since my LAN (and WAN too) are all IPv4; any ideas?

haohaolee:
> Thanks for contributing to Wolf! I'm in the middle of another PR, so it might take me a few days to properly test this out. I saw you've added some unit tests for the basic methods added, brilliant job!
> 
> I was wondering how I can test this out since my LAN (and WAN too) are all IPv4; any ideas?

I guess you can try [Link-Local Address](https://en.wikipedia.org/wiki/Link-local_address?useskin=vector) (those start with `fe80::`) in you LAN, but I have not got a chance to give this a try.

haohaolee:
any update on this? Or need more input?

ABeltramo:
Sorry I've been busy working on the wolf-ui PR and I've had no time to look into this. I'll probably merge it there first since we are going to ask users to beta test all the new changes that are introduced!

haohaolee:
Hi @ABeltramo where can I get the feedback for this beta testing?

ABeltramo:
Sorry, I've been sidetracked by a few other more pressing things but I haven't forgotten about this.
 
I'll probably merge this into `wolf-ui` which is the current WIP dev branch so that we can test it out better before releasing it into `stable`. I'll give it a quick spin over the next few days and I'll give you some feedback.

ABeltramo:
I just gave this a quick test and there a few issues:
1. `.to_v4()` has been deprecated for quite some time, and as a result it's completely missing from Boost on my dev machine (`1.88`). I've tried using 
```c++
boost::asio::ip::make_address_v4(addr.to_v6().to_string());
```
but I'm not sure whether that achieves the same goal.

2. Some tests that you've added are failing on my machine: 
image

```
wolf/tests/platforms/linux/network.cpp:169: Failure:
due to unexpected exception with message:
  Failed to create dummy interface: test_mapped

wolf/tests/platforms/linux/network.cpp:197: Failure:
  REQUIRE_THAT(ipv6_result, Equals(ipv4_result))
with expansion:
  "00:00:00:00:00:00" equals: "2a:64:51:44:c1:b5"

wolf/tests/platforms/linux/network.cpp:197: Failure:
  REQUIRE_THAT(ipv6_result, Equals(ipv4_result))
with expansion:
  "00:00:00:00:00:00" equals: "2a:64:51:44:c1:b5"
```

 3. Ultimately, on my IPv4 network things aren't working. The very first step of reaching `http://<IP>:47989/serverinfo` fails. On localhost, when using `curl` to test this I can see that `curl -v http://127.0.0.1:47989/serverinfo` fails but if I use `localhost`:
```
curl -v http://localhost:47989/serverinfo
* Host localhost:47989 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:47989...
* Connected to localhost (::1) port 47989
* using HTTP/1.x
> GET /serverinfo HTTP/1.1
> Host: localhost:47989
> User-Agent: curl/8.15.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Length: 2012
< 
```
it seems to resolve localhost to the IPv6 version and the request then works correctly. Which begs the question: why are you forcing all addresses to be IPv6 even if the host isn't using it?

haohaolee:
> I just gave this a quick test and there a few issues:
> 
> 1. `.to_v4()` has been deprecated for quite some time, and as a result it's completely missing from Boost on my dev machine (`1.88`). I've tried using
> 
> ```c++
> boost::asio::ip::make_address_v4(addr.to_v6().to_string());
> ```
> 
> but I'm not sure whether that achieves the same goal.

you are right. It's been while since I used boost.asio last time, I may be not aware of all the changes. But I can manage to update this PR.

I think
```c++
boost::asio::make_address_v4(boost::asio::ip::v4_mapped, addr);
```
is enough
> 
> 2. Some tests that you've added are failing on my machine:
> 
> image
> ```
> wolf/tests/platforms/linux/network.cpp:169: Failure:
> due to unexpected exception with message:
>   Failed to create dummy interface: test_mapped
> 
> wolf/tests/platforms/linux/network.cpp:197: Failure:
>   REQUIRE_THAT(ipv6_result, Equals(ipv4_result))
> with expansion:
>   "00:00:00:00:00:00" equals: "2a:64:51:44:c1:b5"
> 
> wolf/tests/platforms/linux/network.cpp:197: Failure:
>   REQUIRE_THAT(ipv6_result, Equals(ipv4_result))
> with expansion:
>   "00:00:00:00:00:00" equals: "2a:64:51:44:c1:b5"
> ```

Could you please provide more info about these tests, e.g. why creating dummy interface failed. Actually I use root to do this test by creating a real interface in Linux to get a unique MAC address
> 
> 3. Ultimately, on my IPv4 network things aren't working. The very first step of reaching `http://<IP>:47989/serverinfo` fails. On localhost, when using `curl` to test this I can see that `curl -v http://127.0.0.1:47989/serverinfo` fails but if I use `localhost`:
> 
> ```
> curl -v http://localhost:47989/serverinfo
> * Host localhost:47989 was resolved.
> * IPv6: ::1
> * IPv4: 127.0.0.1
> *   Trying [::1]:47989...
> * Connected to localhost (::1) port 47989
> * using HTTP/1.x
> > GET /serverinfo HTTP/1.1
> > Host: localhost:47989
> > User-Agent: curl/8.15.0
> > Accept: */*
> > 
> * Request completely sent off
> < HTTP/1.1 200 OK
> < Content-Length: 2012
> < 
> ```
> 
> it seems to resolve localhost to the IPv6 version and the request then works correctly. Which begs the question: why are you forcing all addresses to be IPv6 even if the host isn't using it?
Could you please try the command
`ss -lnt -e | grep 47989`
in this case, I want to know if this socket is an ipv6-only socket or dual-stack socket

To answer your question: the idea behind this is I thought I had 2 options to implement ipv6:
1. create 2 sockets for each service, ipv4 and ipv6, the logic is very clean and straightforward, but the drawback is I need to change a lot:
    - have an option to enable ipv6 or not
    - create an additional ipv6 socket when needed for each service, as well as some additional threads or event loops
2. leverage the functionality called dual-stack socket provided by the OS, one socket to accept both ipv6 & ipv4, the benefit is we don't have to change the code a lot, just replace every "0.0.0.0" with "::", and AKAIK, enet has already made use of this (when creating ipv6 sockets). The magic is the socket option `IPV6_V6ONLY=0`. The cost of this approach is:
    - every address looks like an ipv6 address even it is ipv4 under the hood (::ffff:127.0.0.1), we need to do more handling when formatting addresses
    - In Linux, a dual-socket is the default behavior, but on other OSes, we need to enable it explicitly (maybe)
    - last but not the least, on some Linux devices that does not have ipv6 builtin, this does not work

But I don't think this is the cause why it does not work for your ipv4-only network, modern Linux kernels always have ipv6 stack out there, you can still see many private ipv6 addresses on your interfaces.

Last word, I need to admit I have thought much about these, and had written much code to detect the OS abilities about ipv6, but finally I gave up, and I want to KISS for now. I don't know if many people have ipv4 only kernels, or even run wolf on other OSes. But I think we can make listening on "::" or "0.0.0.0" an option or a config entry for the end user, which should be better.

ABeltramo:
I'll test it out a bit more when I have a bit of free time available. I agree on the KISS and there's probably no need to overcomplicate things.

> But I think we can make listening on "::" or "0.0.0.0" an option or a config entry for the end user, which should be better.

This sounds like the right approach, it can probably be an env variable. If it's an IPv4 address we keep the good old code path, if not, we switch to the new methods.

Finkregh:
If someone wants to test this and doesn’t have modern IP: https://tunnelbroker.net/
You can even follow through with https://ipv6.he.net/certification/ and - at least some years ago - get a free T-shirt ;)

PR #282, opened 2025-10-28
Add Podman / Quadlet setup example

scottsweb:
Covers AMD, Intel and Nvidia. 

One thing to note is that I am using `:wolf-ui` in these examples.

### DIFF
--- a/docs/modules/user/pages/quickstart.adoc
+++ b/docs/modules/user/pages/quickstart.adoc
@@ -508,6 +508,7 @@ Podman Quadlets::
 --
 
 The easiest way to run Wolf is to use a `rootful` service, this isn't a strict requirement; if you give access to the `/dev/dri` devices, `/etc/wolf` and install the required udev rules everything should work.
+
 Place your `wolf.container` file (populated with either the Nvidia or Intel/AMD examples below) in `/etc/containers/systemd`, then run `sudo systemctl daemon-reload` and `sudo systemctl start wolf.service` to start Wolf.
 
 **Nvidia**
@@ -537,7 +538,7 @@ sudo docker volume ls | grep nvidia-driver
 local     nvidia-driver-vol
 ....
 
-3.{empty} This is an additional step for Podman. Start the temporary container you just created. `sudo podman ps -a` will show a contianer using a temporary name (e.g. `musing_pascal`) and your local `gow/nvidia-driver` image. Only once this container is started will your new volume be populated. Run:
+3.{empty} This is an additional step for Podman. Start the temporary container you just created. `sudo podman ps -a` will show a container using a temporary name (e.g. `musing_pascal`) and your local `gow/nvidia-driver` image. Only once this container is started will your new volume be populated. Run:
 
 [source,bash]
 ....
@@ -568,7 +569,6 @@ After=network-online.target podman.socket
 
 [Service]
 TimeoutStartSec=900
-ExecStartPre=-/usr/bin/mkdir -p /tmp/sockets
 ExecStartPre=-/usr/bin/podman rm --force WolfPulseAudio
 Restart=on-failure
 RestartSec=5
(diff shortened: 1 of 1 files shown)


docs/modules/user/pages/quickstart.adoc
```diff
@@ -501,6 +501,172 @@ And if you have a multi GPU setup, don't forget to set the below env variable
...
+RestartSec=5
+StartLimitBurst=5
+
+[Container]
+AutoUpdate=registry
+ContainerName=%N
+HostName=%N
+Image=ghcr.io/games-on-whales/wolf:stable
+AddCapability=CAP_SYS_PTRACE
+AddCapability=CAP_NET_ADMIN
```
ABeltramo: Are those two extra capabilities actually needed?

docs/modules/user/pages/quickstart.adoc
```diff
@@ -501,6 +501,172 @@ And if you have a multi GPU setup, don't forget to set the below env variable
...
+[source,bash]
+....
+[Unit]
+Description=Wolf / Games On Whales
+Requires=network-online.target podman.socket
+After=network-online.target podman.socket
+
+[Service]
+TimeoutStartSec=900
+ExecStartPre=-/usr/bin/mkdir -p /tmp/sockets
```
ABeltramo: This shouldn't be required anymore, we use custom volumes instead of mounting that path from the host

docs/modules/user/pages/quickstart.adoc
```diff
@@ -501,6 +501,172 @@ And if you have a multi GPU setup, don't forget to set the below env variable
...
+AddDevice=/dev/nvidia-caps/nvidia-cap1
+AddDevice=/dev/nvidia-caps/nvidia-cap2
+AddDevice=/dev/nvidiactl
+AddDevice=/dev/nvidia0
+AddDevice=/dev/nvidia-modeset
+Volume=nvidia-driver-vol:/usr/nvidia
+Volume=/dev/:/dev/:rw
+Volume=/run/udev:/run/udev:rw
+Volume=/etc/wolf/:/etc/wolf:z
+Volume=/run/podman/podman.sock:/var/run/docker.sock:ro
```
ABeltramo, line 599: We should probably add a line or two explaining that `/run/podman/podman.sock` might be missing, possibly linking to https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md

ABeltramo (commented):
Thanks for taking the time to write this down!   

I've pushed a small edit replacing `wolf-ui` with `stable` (there's nothing wolf-ui specific and it's going to be merged soon-ish anyway) and I've added a few comments.  
Let me know what you think!

docs/modules/user/pages/quickstart.adoc
scottsweb: They don't seem to be, I have removed them from the examples

docs/modules/user/pages/quickstart.adoc
scottsweb: Removed this also

docs/modules/user/pages/quickstart.adoc
scottsweb: Added some additional pointers for starting the socket

scottsweb:
Updated, thanks for the feedback! 

I am also curious to find out that this should all work `rootless` ... based on previous discussion in Discord I had assumed it had to be `rootful`. I will need to get around to testing that in future.

ABeltramo:
There's been mentions of cgroup rules not being allowed in podman when running rootless and I've found [this discussion](https://github.com/containers/podman/issues/25365). Not sure if that's still the case and if by just mounting `/dev` we can sidestep that issue. At the end of the day we just use that to access our virtual joypads; if that's the only thing stopping us from being able to run fully rootless on podman then I'm sure we can find some way around it..

And FYI when I develop Wolf I just run the executable on my host with the default non-root user. As long as the permissions on the required devices/files are set things should work!

PR #287, opened 2025-11-07
fix: get CUDA device index in ascending order based on PCI addresses

zexi:
Fix the issue: https://github.com/games-on-whales/wolf/issues/233#issuecomment-3471903432

I discovered that Wolf retrieves the cuda_device_id by reading the 'Device Minor' number, which appears to be the source of the problem. On my system, cuda_device_id shoud return in ascending order based on PCI addresses.

image
image


ABeltramo:
Unfortunately I don't have a multi Nvidia setup myself to properly test this but it looks good to me and things still work on my dev machine with a dual AMD/Nvidia GPU so I'm going to merge it. Thanks for contributing!

PR #289, opened 2025-11-15
enhancement: Update API version from 1.40 to 1.50

Matszwe02:
Docker recently pushed an update that doesn't allow the current version to work. API is basically intact, so a simple version rewrite is sufficient.


ABeltramo:
Thanks for taking the time to contribute to Wolf!

Unfortunately this has already been fixed in https://github.com/games-on-whales/wolf/commit/abe0236781e496ad132491e9b1e01aff2d0061c9 so that instead of hard coding a version we automatically pick the current Docker version. This has the benefit that will not break older installations that don't expose v1.50 and should be more future proof.  
I'm going to merge `wolf-ui` fairly soon since it's practically ready to go so I'm going to close this now

PR #305, opened 2025-12-03
Pause/Resume containers

Murazaki:
Attempt at enabling pause/resume scheme from Wolf to Wolf UI.

As discussed on Discord, this might not work perfectly has pausing containers doesn't free GPU VRAM, since there's no cgroup freeze method available to do so by default. There is hope that we could do it with some research, and it might come to the kernel in the next years, but for now we have to try without. This might end up creating crashes though.

I'm still not sure of the return values that `Resume` should give. I kept it as a separated action from `Join` for now, but maybe the scheme would be better if we could directly join ?

I also might have connected pause and resume stream events to pause/resume container, which depending on how this works might not be a good idea.

Linked to https://github.com/games-on-whales/wolf-ui/pull/10

### DIFF
--- a/src/moonlight-server/api/unix_socket_server.cpp
+++ b/src/moonlight-server/api/unix_socket_server.cpp
@@ -282,6 +282,28 @@ UnixSocketServer::UnixSocketServer(boost::asio::io_context &io_context,
 
   state_->http.add(
       HTTPMethod::POST,
+      "/api/v1/lobbies/pause",
+      {
+          .summary = "Pause a lobby",
+          .request_description = APIDescription{.json_schema = rfl::json::to_schema<events::PauseLobbyEvent>()},
+          .response_description = {{200, {.json_schema = rfl::json::to_schema<GenericSuccessResponse>()}},
+                                   {500, {.json_schema = rfl::json::to_schema<GenericErrorResponse>()}}},
+          .handler = [this](auto req, auto socket) { endpoint_LobbyPause(req, socket); },
+      });
+
+  state_->http.add(
+      HTTPMethod::POST,
+      "/api/v1/lobbies/resume",
+      {
+          .summary = "Resume a lobby",
+          .request_description = APIDescription{.json_schema = rfl::json::to_schema<events::ResumeLobbyEvent>()},
+          .response_description = {{200, {.json_schema = rfl::json::to_schema<GenericSuccessResponse>()}},
+                                   {500, {.json_schema = rfl::json::to_schema<GenericErrorResponse>()}}},
+          .handler = [this](auto req, auto socket) { endpoint_LobbyResume(req, socket); },
+      });
+
+  state_->http.add(
+      HTTPMethod::POST,
       "/api/v1/lobbies/stop",
       {
           .summary = "Stop a lobby",
--- a/src/moonlight-server/api/unix_socket_server.cpp
+++ b/src/moonlight-server/api/unix_socket_server.cpp
@@ -101,36 +101,38 @@ UnixSocketServer::UnixSocketServer(boost::asio::io_context &io_context,
       });
 
   state_->http.add(HTTPMethod::POST,
-                   "/api/v1/apps/add",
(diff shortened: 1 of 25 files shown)


src/moonlight-server/runners/docker.cpp
```diff
@@ -196,6 +196,34 @@ void RunDocker::run(std::string_view session_id,
     logs::log(logs::info, "[DOCKER] Starting container: {}", docker_container->name);
     logs::log(logs::debug, "[DOCKER] Starting container: {}", *docker_container);
 
+    auto pause_handler = this->ev_bus->register_handler<immer::box<events::PauseStreamEvent>>(
+        [session_id, container_id, this](const immer::box<events::PauseStreamEvent> &pause_ev) {
+          if (std::to_string(pause_ev->session_id) == session_id) {
+            docker_api.pause_by_id(container_id);
+          }
+        });
```
ABeltramo, line 204: Don't handle `PauseStreamEvent` (and`ResumeLobbyEvent` below) . For single users this can potentially break apps and for lobbies it will pause an entire lobby when any participant exits which you definitely don't want to happen.

src/moonlight-server/sessions/lobbies.cpp
```diff
@@ -253,6 +253,39 @@ setup_lobbies_handlers(const immer::box<state::AppState> &app_state,
...
+        auto lobbies = app_state->lobbies->load();
+        auto lobby = state::get_lobby_by_id(lobbies.get(), resume_lobby_event->lobby_id);
+
+        if (!lobby) {
+          logs::log(logs::error,
+                    "[LOBBY] Failed to resume lobby: lobby {} not found",
+                    resume_lobby_event->lobby_id);
+          return;
+        }
+      }));
```
ABeltramo: This doesn't do anything. I think it can be safely removed.

src/moonlight-server/events/events.hpp
```diff
@@ -192,6 +192,16 @@ struct LeaveLobbyEvent {
...
 
+struct PauseLobbyEvent {
+  const std::string lobby_id;
+  std::optional<std::vector<short>> pin = std::nullopt;
+};
+
+struct ResumeLobbyEvent {
+  const std::string lobby_id;
+  std::optional<std::vector<short>> pin = std::nullopt;
+};
```
ABeltramo: I think the names here are misleading, I'd use something like `PauseRunner` and `ResumeRunner` instead. In fact, with your implementation, when the pause gets called the lobby is effectively destroyed (since everyone leaves).

ABeltramo (changes requested):
Thanks for working on this, looks good! Just a couple of notes and small changes from my side.

Also, I would prefer to remove all the extra `.envrc` and `nix` files if you can.

src/moonlight-server/api/unix_socket_server.cpp
```diff
@@ -280,6 +280,28 @@ UnixSocketServer::UnixSocketServer(boost::asio::io_context &io_context,
...
+      HTTPMethod::POST,
+      "/api/v1/lobbies/resume",
+      {
+          .summary = "Resume a lobby",
+          .request_description = APIDescription{.json_schema = rfl::json::to_schema<events::ResumeLobbyEvent>()},
+          .response_description = {{200, {.json_schema = rfl::json::to_schema<GenericSuccessResponse>()}},
+                                   {500, {.json_schema = rfl::json::to_schema<GenericErrorResponse>()}}},
+          .handler = [this](auto req, auto socket) { endpoint_LobbyResume(req, socket); },
+      });
+
```
ABeltramo: This also should be `/api/v1/runners/pause` and `/api/v1/runners/resume`

ABeltramo:
Also the following condition 

```c++
while (docker_api.get_by_id(container_id)->status == RUNNING)
```

should probably be changed: when we get out of that loop, we'll stop the container and trigger the chain of events that will end the lobby.

PR #307, opened 2025-12-04
Fix podman command in quadlet quickstart guide

iKadmium:
This fixes one of the commands in the podman quadlet quickstart guide which refers to docker instead of podman


ABeltramo:
Thanks for contributing!

PR #343, opened 2026-01-23
fix: gstreamer build option (nvcodec)

jbdemonte:


ABeltramo:
Thanks for contributing, not sure how I've missed that, but I suppose the `nvcodec` is enabled by default anyway 😅

PR #347, opened 2026-02-04
Add Pulse Audio Router to ensure correct sink is selected

### LINKED ISSUE #269: Audio of some games/apps goes to the wrong sink
When launching some games, notably, Victoria 3 and Minecraft, no game audio can be heard in the session in which they are launched.

Instead, the game audio is audible in the wolf-ui main menu.

**Workarounds:**

- Some affected games, such as Minecraft, allow the user to select an alternate audio sink for game audio. Changing the audio sink to "Null Output <span>#</span>2" solves this issue.
- Another workaround is disconnecting the Moonlight client from the Wolf server while leaving the Steam/PrismLauncher/etc. session running, quitting the Wolf UI session, restarting the Wolf UI session, and re-entering the still running Steam/PrismLauncher/etc. session. At this point, all games should be audible inside the proper session.

Please let me know if you need any more information. Thank you.

camelcx:
Implemented a pulse router to ensure correct sound input -> pulse sink routing. Some APPs ignore the $PULSE_SINK env and choose the wrong sink to output sound breaking multiuser isolation.

### DIFF
--- a/src/moonlight-server/audio/pulse_router.cpp
+++ b/src/moonlight-server/audio/pulse_router.cpp
@@ -43,6 +43,11 @@ setup_pulseaudio_router_handlers(const immer::box<state::AppState>& app_state,
         if (state) state->on_container_stopped(*ev);
       }));
 
+  handlers.push_back(app_state->event_bus->register_handler<immer::box<events::VirtualAudioSinkCreated>>(
+      [state](const immer::box<events::VirtualAudioSinkCreated>& ev) {
+        state->on_virtual_sink_created(*ev);
+      }));
+
   // Subscribe to sink-input events and do an initial scan
   state->enable_pulse_subscribe();
 
@@ -83,15 +88,45 @@ void PulseAudioRouterState::on_container_created(const events::DockerContainerCr
 void PulseAudioRouterState::on_container_stopped(const events::DockerContainerStopped& ev) {
   if (ev.hostname.empty()) return;
 
+  std::optional<std::string> removed_session;
+
   host_to_session.update([&](auto m) {
-    return m.erase(ev.hostname);
+    if (auto v = m.find(ev.hostname)) {         // hostname -> session_id
+      // optionally guard against hostname reuse:
+      if (ev.session_id.empty() || *v == ev.session_id) {
+        removed_session = *v;
+        return m.erase(ev.hostname);
+      }
+    }
+    return m;
+  });
+
+  if (removed_session && !removed_session->empty()) {
+    session_to_sink_idx.update([&](auto m) { return m.erase(*removed_session); });
+
+    logs::log(logs::debug,
+              "[PULSE_ROUTER] Removed mappings host='{}' session='{}'",
+              ev.hostname,
+              *removed_session);
+  }
+
+}
+
+void PulseAudioRouterState::on_virtual_sink_created(const events::VirtualAudioSinkCreated& ev) {
+  if (ev.session_id.empty()) return;
+
+  session_to_sink_idx.update([&](auto m) {
+    return m.set(ev.session_id, ev.sink_index);
   });
 
   logs::log(logs::debug,
-            "[PULSE_ROUTER] Map del host='{}' session='{}' (container_id='{}')",
-            ev.hostname,
+            "[PULSE_ROUTER] Sink ready session='{}' sink='{}' idx={}",
             ev.session_id,
-            ev.container_id);
+            ev.sink_name,
+            ev.sink_index);
+
+  // Fix races: sink-input may already exist, or docker mapping may already exist
+  rescan();
 }
 
 void PulseAudioRouterState::enable_pulse_subscribe() {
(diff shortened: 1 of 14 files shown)


src/moonlight-server/audio/pulse_router.cpp
```diff
@@ -0,0 +1,175 @@
...
+  });
+}
+
+void PulseAudioRouterState::on_container_created(const events::DockerContainerCreated& ev) {
+  if (ev.hostname.empty() || ev.session_id.empty()) return;
+
+  {
+    std::lock_guard<std::mutex> lk(map_mu);
+    host_to_session[ev.hostname] = ev.session_id;
+  }
```
ABeltramo: To keep the codebase consistent it would be better to use 
```c++
immer::map<std::string, std::string>
```
instead of having to manually deal with the mutex

src/moonlight-server/wolf.cpp
```diff
@@ -223,6 +224,10 @@ void run() {
   }).detach();
 
   auto audio_server = setup_audio_server(local_state->host->host_xdg_runtime_dir, runtime_dir);
+  // PulseAudio sink-input router (hostname -> session_id -> virtual_sink_<session>)
+  auto pulse_router_state = std::make_shared<audio::PulseAudioRouterState>();
+  pulse_router_state->pulse_server = audio_server->server;
+  auto pulse_router_handlers = audio::setup_pulseaudio_router_handlers(local_state, pulse_router_state);
```
ABeltramo: I feel that `audio_server->server` should be passed in the `PulseAudioRouterState` constructor, or am I missing something?

src/moonlight-server/audio/pulse_router.hpp
```diff
@@ -0,0 +1,60 @@
...
+  std::shared_ptr<audio::Server> pulse_server;
+  std::string sink_prefix = "virtual_sink_";
+
+  // hostname -> session_id
+  mutable std::mutex map_mu;
+  std::unordered_map<std::string, std::string> host_to_session;
+
+  // Pulse subscribe / callbacks
+  void enable_pulse_subscribe();
+  void disable_pulse_subscribe();
```
ABeltramo: Not implemented, should be removed

src/moonlight-server/audio/pulse_router.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+struct Server;
+
+/**
+ * PulseAudio routing state.
+ * Kept alive by the caller together with the returned handler registrations.
+ */
+struct PulseAudioRouterState {
+  std::shared_ptr<audio::Server> pulse_server;
+  std::string sink_prefix = "virtual_sink_";
```
ABeltramo: This hardcoded string stinks, and now that I look at the code it's hardcoded in a lot of places..  
Might be better to make it a `constexpr std::string_view` in this header and just reference that from all the places where it's currently hardcoded.

ABeltramo (changes requested):
Thanks for working on this, it looks really clean!  
Just a few minor changes and nitpicks here and there; let me know what you think!

src/moonlight-server/audio/pulse_router.cpp
camelcx: Done in commit c0389a9.

src/moonlight-server/wolf.cpp
camelcx: Nope, you are not missing anything. changed to pass in constructor. Commit 71278d6

src/moonlight-server/audio/pulse_router.hpp
camelcx: Removed now.

src/moonlight-server/audio/pulse_router.hpp
camelcx: Working on this one. I think it might be better having this string defined close to where the virtual sink was created and everything reference to that. I'll see what I can do.

camelcx: Created a common.hpp under sessions just to store this. Managed to create a compilation friendly typo that cuts most the audio off... Took me a while to fix. Now it's good.

camelcx:
The current implementation reroutes blindly whether or not the input is routed to the correct sink or not. It works, but sometimes does quite a few reroute attempts even if the input is already on the correct sink. Ideally, we would check first and only reroute when the existing connection is wrong. However, the input info we get inside the subscribe call back do not contain the sink name but only the index. We can probably fire some new events like VirtualSinkCreated and build another map of sink_index->sink_name (since the index is available on creation)? 

Alternatively, we can get the index info by pa_context_get_sink_info_by_name(). But that is async and we need to guard against race condition... @ABeltramo

camelcx:
Implemented the event and checks as above. All works fine for me now. The index of VSink (return value from create_virtual_sink call) turns out to be the owner module index, which is different from the actual sink index for audio routing. Had to put a workaround there...

linxuhao:
I'm just here to say thank you !

camelcx:
@linxuhao you are welcome. Glad it helps :)

ABeltramo (approved):
I've tested it a bit running a few different sessions trying both to isolate sound and co-op and everything seems to work quite smoothly. Amazing job, thanks for contributing!

PR #349, opened 2026-02-05
note that volume with old drivers should be removed

undefx:
context: https://discord.com/channels/856434175455133727/1468737628303261831


ZackaryH8:
Could wolf not check if there is a mismatch or older version and actually display a meaningful error message in the logs?

PR #361, opened 2026-02-18
Documentation/Quickstart: Enhance lxc quickstart guide with GPU driver notes for nvidia

DerPfandadler:
Added note about GPU driver device creation and suggested running nvidia-smi at boot for nvidia gpu Proxmox lxc hosts.

### DIFF
--- a/docs/modules/user/pages/quickstart.adoc
+++ b/docs/modules/user/pages/quickstart.adoc
@@ -398,6 +398,21 @@ At the moment it is only possible to run Wolf inside a privileged LXC.
 
 First you need to make sure your GPU drivers are installed and loaded on your PVE host.
 
+[NOTE]
+====
+On some systems the `/dev/nvidia*` devices are only created after the driver is first used.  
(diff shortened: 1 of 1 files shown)


DerPfandadler:
I just created a tiny oneshot service for that. so the lxc can auto start after bootup. But i found it a bit too simple and long to include right there in the quickstart, right?

```
[Unit]
Description=Run nvidia-smi once at boot to initialize GPU
After=systemd-modules-load.service
Before=pve-guests.service

[Service]
Type=oneshot
ExecStart=/usr/bin/nvidia-smi

[Install]
WantedBy=multi-user.target
```

ABeltramo:
Thanks for opening this PR!  
There's another more specific cli tool that just creates the Nvidia extra devices but I can't recall it from the top of my head.. It's probably buried in the messages in our Discord server somewhere.  

The same suggestion should be applied for all the Nvidia tabs btw.

nhths:
On my Proxmox VE host, I need to execute `nvidia-modprobe -u -m -s -l -c 0` after `nvidia-smi`. Otherwise, some /dev/* will not be initialized.

PR #366, opened 2026-03-09
Fix typo in quickstart.adoc command

Robotathand:
Corrected "sudo la -la ..." to "sudo ls -la ..."


ABeltramo:
Thanks for contributing!

PR #367, opened 2026-03-10
Fix Lutris config for version 0.5.19 or later

clsty:
This PR is a part of another PR [gow#296](https://github.com/games-on-whales/gow/pull/296) which uses Lutris later than 0.5.19 (currently 0.5.22 as latest).

---

A side note: According to my test, `SYS_PTRACE` is necessary when using Gamescope with Proton, else gamescope gives error: `xdg_surface has never been configured`.


ABeltramo:
I'm going to merge this after we've merged https://github.com/games-on-whales/wolf/pull/368 because we are going to upgrade the config file in that PR, and so we can slot in this new default too

clsty:
> I'm going to merge this after we've merged #368 because we are going to upgrade the config file in that PR, and so we can slot in this new default too

`src/moonlight-server/state/default/config.v6.toml` here does not conflict with `src/moonlight-server/state/default/config.v7.toml` which #368 changes. I suggest to merge this PR now (which is necessary for Lutris > 0.5.19 to work normally), and handle it separately about `config.v7.toml`. What do you think?

clsty:
> `src/moonlight-server/state/default/config.v6.toml` here does not conflict with `src/moonlight-server/state/default/config.v7.toml` which #368 changes. I suggest to merge this PR now (which is **necessary for Lutris > 0.5.19 to work normally with proton**), and handle it separately about `config.v7.toml`. What do you think?

@ABeltramo I'll merge this tomorrow if you don't mind.

ABeltramo:
@clsty please don't.  
Merging this will only "fix it" for new installations, current users wouldn't get the upgraded config. This change should be included and merged with https://github.com/games-on-whales/wolf/pull/368 which, by upgrading the config, will have a path to upgrade the configuration for current users.

clsty:
@ABeltramo Ok, I'm not merging it then. However I'm still wondering why.

> Merging this will only "fix it" for new installations

Agree, that's why I wanted to merge it. I think at least we should always provide a correct default config for a fresh installation, or else a new user may find out Lutris not working with Proton and just think "Hmm what? Lutris image is broken. Wait, maybe it's a configuration problem... No it can't be, as I'm already using the latest default config."

> current users wouldn't get the upgraded config.

Agree, however merging this PR does no harm to them anyway. Also if a WolfLutris user finds their container not working even using the latest image, they'll probably checkout the latest `config.v6.toml`; but if we're not merging this now, then the latest `config.v6.toml` does not help this user.

> This change should be included and merged with https://github.com/games-on-whales/wolf/pull/368 which, by upgrading the config, will have a path to upgrade the configuration for current users.

Correct, however not neccessary, as we can modify PR-368 separately. Merging this PR should not affect the availability of modifying PR-368. And more importantly this PR modifies `config.v6.toml` which is even not the same file as `config.v7.toml` which PR-368 modifies, which means that after merging PR-368 we'll have to manually modify `config.v7.toml` anyway, without available auto magic.

ABeltramo:
All good points, I just feel that this change might easily get lost in #368 since, as you noticed, there's no file conflict (v6 vs v7).  
I think we are also close to merge that one, so just waiting a few days for this to be cleanly applied for both new and existing users in one go should be better if that makes sense.

clsty:
> I just feel that this change might easily get lost in https://github.com/games-on-whales/wolf/pull/368 since, as you noticed, there's no file conflict (v6 vs v7).

I'm also aware of this fact. That's why I suggest to handle PR-368 separately, which specifically means that I can modify PR-368 directly.

I'll modify PR-368 to sync the changes made here, if you agree. Or, I can create an issue to track it.

> I think we are also close to merge that one, so just waiting a few days for this to be cleanly applied for both new and existing users in one go should be better if that makes sense.

Actually I plan to modify PR-368 **together** with merging this PR, so I guess this could be considered as a "one go".

ABeltramo:
#368 has been merged, can you rebase on top of latest `:stable` and add your changes to `config.v7.toml` instead of `v6`? You should also run `cmake` so that it generates `config.include.toml` (which is what actually gets used as default by Wolf)

clsty:
Sure, will get this done when I'm available.

clsty:
I've got it done but these automatically triggered tests failed which looks not very well. I hope it's not related with this PR.

src/moonlight-server/state/default/config.v6.toml
ABeltramo, line 1: This file shouldn't be deleted, can you revert this change?

ABeltramo (changes requested):
Don't worry about GitHub Actions; looks like it's just failing because it's a fork (and they aren't configured correctly)

src/moonlight-server/state/default/config.v6.toml
clsty: Sure, done.

PR #368, opened 2026-03-13
fix: add VBV buffer constraints and improve UDP streaming

### LINKED ISSUE #332: Bitrate settings do not work correctly in Wolf. A bitrate of 0.5 Mb/s in Moonlight yields a output from Wolf between 0.225MB/s up to 11.2MB/s
I'm not sure if this issue was already present since the beginning of Wolf or if it's something recent, but I noticed that the bitrate setting from Moonlight is not used as bitrate ceiling in Wolf. 

When I set a Video Bitrate in Moonlight to the smallest available setting of 0.5MB/s, the resulting stream sits around 0.225MB/s when there is no motion on screen, with a bit of motion it averages around 4-6MB/s, and with a lot of motion, it can go up to 11.2MB/s from my own tests.
This is way over 0.5MB/s. 

The image looks crystal clear with the 0.5MB/s bitrate. I tested with H264, HEVC and AV1, without seeing much difference in bitrate behavior.
With Nvidia Experience, the video looked extremely blocky, but the stream bandwidth would never go above 0.5MB/s as expected. This is what I'd expect here to happen too. This issue makes Wolf unusable over internet within internet speed constrained networks, even on the lowest bitrate setting, the image constantly freezes as the network bandwidth gets constantly exceeded whenever there is a bit more motion in the video.

I was able to observe that when the bitrate is increased in Moonlight, the internet bandwidth used by the Wolf stream also increases (Both for the minimum without motion as as well the maximum with motion), indicating that the bitrate setting itself IS received on Wolf side. 

I completely deleted my "wolf.toml" config file, so the settings are all default, and the same issue is still present. I am using the AMD igpu encoder. Could that be the cause?

Baz00k:
## Summary

Fix streaming instability on clients with limited receive buffers (e.g. Moonlight-TV on embedded devices) and Wolf ignoring client bitrate settings.
Changes in this PR allowed me to go from unstable 12mbit stream with significant packet loss to full, stable 35mbit on my TV. This should also stabilize streaming over Wi-Fi.

**Root cause**: Missing VBV (Video Buffering Verifier) buffer constraints on all hardware encoder pipelines. Without VBV, CBR mode allows individual frames — especially IDR frames and complex game scenes — to spike 2-5x the per-frame bitrate budget, overwhelming client receive buffers.

Most of this code is just reimplementation of what alredy exists in Sunshine/Apollo.

### 1. Apollo-style batched UDP send
- Replace async single-packet `async_send_to()` calls with synchronous batched sends using `sendmmsg()` 
- Add platform abstraction layer (`batched_send.hpp/cpp`) with Linux implementation
- Configure socket QoS and buffer sizes optimized for streaming workloads
- Increase default video packet payload size to match Moonlight protocol spec

### 2. Packet pacing
- Add rate-limited packet pacing for video UDP sends (80% of 1Gbps link speed)
- Sub-millisecond sleep-based scheduling between send windows
- Smooths out large frame bursts that would could otherwise overwhelm low-powered devices

### 3. VBV buffer constraints for all encoders
- **NVIDIA (nvcodec)**: Add `vbv-buffer-size` to all `nvh265enc`, `nvh264enc`, `nvav1enc` pipelines (config v3-v6 + include)
- **VAAPI (va)**: Fix `rate-control` from `cqp` (Constant QP — ignores bitrate entirely) to `cbr`, and fix `cpb-size` from `{bitrate}` (1-second buffer) to single-frame buffer
- VBV buffer size computed as `bitrate_kbps / refreshRate` (one frame's worth), matching Sunshine/Apollo's approach

## Related issues

Fixes #332 

## Testing

Tested on Moonlight-TV at 35 Mbps AV1 with stable streaming (previously broke above 12 Mbps).
Streaming still works on original Moonlight and Artemis clients (tested on S23 Ultra and M1 Mac).

## Files Changed

- `streaming/streaming.cpp` — Batched send, pacing, VBV format arg
- `platform/batched_send.{hpp,cpp}` — New batched send abstraction
- `platform/linux/batched_send.cpp` — Linux sendmmsg implementation  
- `gstrtpmoonlightpay_video.cpp`, `rtsp/commands.hpp` — Packet size
- `config.include.toml`, `config.v3-v6.toml` — Encoder pipeline fixes


nhths:
I'm experiencing occasional audio stuttering on my client (Manjaro, Ryzen 5500U laptop), which I suspect is due to insufficient audio buffer size on the sender/receiver side.
Could you clarify if this PR addresses audio buffering at all, or is it strictly limited to video pipeline (VBV/pacing)?

Baz00k:
> I'm experiencing occasional audio stuttering on my client (Manjaro, Ryzen 5500U laptop), which I suspect is due to insufficient audio buffer size on the sender/receiver side. Could you clarify if this PR addresses audio buffering at all, or is it strictly limited to video pipeline (VBV/pacing)?

No, this PR affects video pipeline only and packet pacing is not enabled for audio sink.
Unfortunately, your audio issues require a separate investigation.

Trasen:
I am very open to testing this one out, got a docker build up?

Baz00k:
@Trasen Sure! https://github.com/users/Baz00k/packages/container/package/wolf Only use the latest build, the rest will  not work

Trasen:
Jupp, works great 👍 👌 👏

Trasen:
Actually more than great. In resource strained scenarios it seem to ease the encoder volatility. So if your gpu is pushing 100% the encoder seem to have an easier time.

src/moonlight-server/gst-plugin/gstrtpmoonlightpay_video.cpp
```diff
@@ -102,7 +102,7 @@ static void gst_rtp_moonlight_pay_video_class_init(gst_rtp_moonlight_pay_videoCl
           "Maximum size of RTP packets. If a video payload surpasses this it'll be split in multiple packets",
           0,
           10240,
-          1024,
+          1392,
```
ABeltramo: Where have you taken this value from?

src/moonlight-server/platform/batched_send.cpp
```diff
@@ -0,0 +1,22 @@
...
+          desc.size - offset,
+      };
+    } else {
+      offset -= desc.size;
+    }
+  }
+  return {};
+}
+
+} // namespace wolf::platform
```
ABeltramo: Why is this defined in this extra file?

src/core/src/core/batched_send.hpp
```diff
@@ -0,0 +1,113 @@
+#pragma once
+
+#include <boost/asio/ip/address.hpp>
+#include <boost/asio/ip/udp.hpp>
+#include <cstdint>
+#include <vector>
+
+namespace wolf::platform {
```
ABeltramo, line 8: The way the Wolf codebase is currently layed out it would make more sense to move this in `src/core` instead. We already have the logic for switching implementations between platforms under that `CMakeLists.txt`

src/moonlight-server/rtsp/commands.hpp
```diff
@@ -225,10 +225,10 @@ announce(const RTSP_PACKET &req, const events::StreamSession &session) {
 
       .port = session.video_stream_port,
       .timeout_ms = args["x-nv-video[0].timeoutLengthMs"].value_or(7000),
-      .packet_size = args["x-nv-video[0].packetSize"].value_or(1024),
+      .packet_size = args["x-nv-video[0].packetSize"].value_or(1392),
       .frames_with_invalid_ref_threshold = args["x-nv-video[0].framesWithInvalidRefThreshold"].value_or(0),
       .fec_percentage = fec_percentage,
-      .min_required_fec_packets = args["x-nv-vqos[0].fec.minRequiredFecPackets"].value_or(0),
```
ABeltramo, line 231: Where does this default comes from? Sunshine is defaulting to `0` if omitted IIUC

src/moonlight-server/state/default/config.v3.toml
```diff
@@ -173,7 +173,7 @@ chroma-site={color_range}, format=NV12, colorimetry={color_space}, pixel-aspect-
 \
 """
 encoder_pipeline = """
-nvh265enc preset=low-latency-hq zerolatency=true gop-size=-1 rc-mode=cbr-ld-hq bitrate={bitrate} aud=false !
+nvh265enc preset=low-latency-hq zerolatency=true gop-size=-1 rc-mode=cbr-ld-hq bitrate={bitrate} vbv-buffer-size={vbv_buffer_size} aud=false !
```
ABeltramo: I like to keep the old config versions in order to easily keep trace of old defaults. Please revert the changes in `v3`, `v4`, and `v5`

src/moonlight-server/state/default/config.include.toml
```diff
@@ -375,7 +375,7 @@ video/x-raw(memory:VAMemory), format=NV12, width={width}, height={height}, pixel
 plugin_name = "nvcodec"
 check_elements = ["nvh265enc", "cudaconvertscale", "cudaupload"]
 encoder_pipeline = """
-nvh265enc gop-size=-1 bitrate={bitrate} aud=false rc-mode=cbr zerolatency=true preset=p1 tune=ultra-low-latency multi-pass=two-pass-quarter !
+nvh265enc gop-size=-1 bitrate={bitrate} vbv-buffer-size={vbv_buffer_size} aud=false rc-mode=cbr zerolatency=true preset=p1 tune=ultra-low-latency multi-pass=two-pass-quarter !
```
ABeltramo, line 379: Wolf servers wouldn't pick up these important changes unless we bump up the config version. We should add a new `v7` config and an upgrade path in `configTOML.cpp`

src/moonlight-server/streaming/streaming.cpp
```diff
@@ -191,29 +194,139 @@ void start_audio_producer(const std::string &session_id,
 
 namespace custom_sink {
 
+struct PacingConfig {
+  bool enabled = false;
+  std::size_t max_packets_per_ms = 0;
+  std::size_t max_batch_size = 16;
```
ABeltramo, line 207: Why 16? How have you picked this number?

ABeltramo (changes requested):
Thanks for taking the time to contribute to Wolf!  
I've taken a look at the code and added a few preliminary comments before I fully dive into the new `sendmmsg` implementation..

src/core/src/core/batched_send.hpp
Baz00k: Done!

src/moonlight-server/gst-plugin/gstrtpmoonlightpay_video.cpp
Baz00k: This is a default value used by moonlight clients (https://github.com/moonlight-stream/moonlight-qt/blob/2e9fbecfea388ba762ffce93ceaecc6d76f9fbba/app/streaming/session.cpp#L1658)

src/moonlight-server/rtsp/commands.hpp
Baz00k: Sunshine does default this to 0 but in practice clients overwride this value to 2, as it just makes more sense (https://github.com/moonlight-stream/moonlight-common-c/blob/62687809b1f7410c3db4be2527503a54ae408d70/src/SdpGenerator.c#L210)

src/moonlight-server/state/default/config.include.toml
Baz00k: Sure, can do. What should be the migration path? Full config replacement or only patch the encoder pipeline strings?

src/moonlight-server/streaming/streaming.cpp
Baz00k: Just an arbitrary value that follows the Sunshine's pattern of capping batch size to under 64KB per `sendmmsg` (https://github.com/LizardByte/Sunshine/blob/ff33670c84e0655eeffff05e40cb846db2fdc487/src/stream.cpp#L1414). 

It is calculated at runtime on line 404: ` std::min<std::size_t>(16, 65536 / video_session->packet_size),` ( so with the default packet size of 1392 it evaluates to `min(16, 47) = 16)`), which means that the value could be slightly increased to reduce the number of syscalls.

Does not affect pacing anyway.

src/moonlight-server/gst-plugin/gstrtpmoonlightpay_video.cpp
ABeltramo: Perfect, I'd suggest adding that in a comment in that line then.

src/moonlight-server/rtsp/commands.hpp
ABeltramo: Same as above, adding that as a comment for posterity would be good

src/moonlight-server/streaming/streaming.cpp
ABeltramo: Makes sense, here in the `struct` we can add a proper docstring with `/**` above that param :wink:

src/moonlight-server/state/default/config.include.toml
ABeltramo: Ideally we would do the inverse of https://github.com/games-on-whales/wolf/blob/5a393daafac36ff86453504d96faea50d160780d/src/moonlight-server/state/configTOML.cpp#L220-L243
something like:
```c++
if (version <= 6) {
  // load up old version and rename to `.v6.old`
  // Replace `gstreamer.video` encoders with the new defaults
  // save the new version
}
```

src/moonlight-server/streaming/streaming.cpp
```diff
@@ -270,26 +375,34 @@ void start_streaming_video(immer::box<events::VideoSession> video_session,
...
+      fmt::arg("color_range", color_range),
+      fmt::arg("host_port", video_session->port));
   logs::log(logs::debug, "Starting video pipeline: \n{}", pipeline);
 
   std::shared_ptr<custom_sink::UDPSink> udp_sink = std::make_shared<custom_sink::UDPSink>(custom_sink::UDPSink{
       .socket = video_socket,
-      .client_endpoint = std::make_shared<udp::endpoint>(boost::asio::ip::make_address(client_ip), client_port)});
+      .client_endpoint = std::make_shared<udp::endpoint>(boost::asio::ip::make_address(client_ip), client_port),
+      .pacing = {
+          .enabled = true,
```
ABeltramo: I'd suggest making this a simple ENV variable something like `WOLF_ENABLE_VIDEO_PACING` so that we can easily toggle this new behaviour when users start running this in the wild..

Baz00k:
@ABeltramo All comments addressed, can you take a look again?

ABeltramo:
I was trying to push a commit, but it looks like you haven't toggled the flag to give contributors permissions. I'll add a few more comments..

src/moonlight-server/state/default/config.v7.toml
```diff
@@ -0,0 +1,559 @@
...
+
+default_opus_encoder = """
+opusenc bitrate={bitrate} bitrate-type=cbr frame-size={packet_duration} bandwidth=fullband \
+audio-type=restricted-lowdelay max-payload-size=1400\
+"""
+
+default_sink = """
+rtpmoonlightpay_audio name=moonlight_pay packet_duration={packet_duration} encrypt={encrypt} \
+aes_key="{aes_key}" aes_iv="{aes_iv}" !
+appsink name=wolf_udp_sink\
```
ABeltramo, line 561: You are missing the closing `"""` here

src/moonlight-server/state/default/config.include.toml
```diff
@@ -557,5 +557,4 @@ audio-type=restricted-lowdelay max-payload-size=1400\
 default_sink = """
 rtpmoonlightpay_audio name=moonlight_pay packet_duration={packet_duration} encrypt={encrypt} \
 aes_key="{aes_key}" aes_iv="{aes_iv}" !
-appsink name=wolf_udp_sink\
-""")for_c++_include"
\ No newline at end of file
+appsink name=wolf_udp_sink\)for_c++_include"
```
ABeltramo: And this should be regenerated with cmake

ABeltramo:
Also tests are failing because the test config file is still at version `6`:
https://github.com/Baz00k/wolf/blob/99aa9e74c6f16de699134ec186ef9d3a8870bd49/tests/assets/config.test.toml#L9

ABeltramo:
On a more general note, I've given this branch a go playing a game for a bit and I can see that the new encoding pipeline config is much better at constraining the bitrate set.  
Looks like there's a **memory leak** though, you can quite easily see that the memory used keeps increasing whilst streaming (which wasn't the case on `:stable`).. :thinking:

Baz00k:
@ABeltramo Did some testing and I too noticed an increased memory usage during streaming. For me, it was about 80MB during an hour long test. I didn't find any memory leaks in the code so I suspected that it might be an unbounded appsink queue growing and added `max-buffers=1 drop=true` to the new config. Also, did some small optimizations to the packet pacing code. After those changes I conducted 2 more tests with additional logging and did not find anything suspicious. The memory does still grow initially, but it stabilizes after a few minutes and in my tests never reached above 675MB for the wolf container.
At this point, I do not have evidence that the new UDP batching/pacing code is leaking or causing queue accumulation.

Please let me know if you find anything interesting.

Trasen:
Tried the latest version, you know there is a specific type of tree in Oblivion: Remastered which if i run :Stable and go close to it - the video stream freezes until I look away from it. This is ofc on 4K60/120

That specific issue got resolved by the latest pull. Did you add env variable to activate your specific changes btw? If so, is there something you changed that is not behind that feature toggle?

Baz00k:
@Trasen There is env var `WOLF_ENABLE_VIDEO_PACING` to optionally disable packet pacing. All changes should be enabled by default. Of course you can still manually change video pipelines in your config if you want.

DanielRiek:
I was using CBR for a long time now. Sometimes I still had hickups on my devices in heavy scenes (100Mbit HEVC AMD 9070 XT). Especially with maximum GPU utilization. This was almost completely solved by this PR.  Most recent example: Death Stranding 2 (4k, max details, FSR4 AA). Looks like the encoder is more stable on max utilization of GPU/CPU.

aakister:
I played for a few hours today using the latest build and didn't encounter any issues. AMD GPU and Intel CPU on unraid.

ABeltramo:
I've tested this a bit more, and even leaving aside the memory leak I see **a massive performance hit** when using this, here's an `nvtop` screenshot when running `stable` at 4K:

Screenshot_20260330_195916

Here's a screenshot of running the latest commit in this branch at the same resolution on the exact same HW:

Screenshot_20260330_200128

Something is definitely wrong, is it just my machine tripping me?

Obviously this is less evident at lower resolutions, which makes me think that we are copying the buffers and possibly leaking data somewhere..  

The interesting part though is the massive difference in **GPU usage** which I really can't explain. I've tried reverting to `cbr` but that doesn't seem to affect things. I've tried both an AMD GPU and a Nvidia GPU (on the same machine and OS) and both seem to be producing similar differences with current `stable`.   
I'm running Arch, so I'm using latest Gstreamer `1.28.1`, I'll try building the Docker images to see if I can reproduce the issue with what we bundle. :thinking:

ABeltramo:
We are discussing this on Discord if anyone wants to join https://discord.com/channels/856434175455133727/1488255788219764828 it seems that @DanielRiek too is experiencing a small memory leak when streaming at `4k@60 - 100mbit`. Roughly 1MB/minute of streaming.

ABeltramo:
I've pushed a few small tweaks, here's a quick recap of my findings:
- passing from `cqp` to `cbr` increases the buffer size
- this causes higher CPU usage (to be expected)
- which can be mitigated by adding a `queue` in front

**Anybody testing this before the latest commits** I've pushed should change the `config.toml` to add `queue ! ` before `rtpmoonlightpay_video` as in 
```toml
[gstreamer.video]

default_sink = """
queue ! \
rtpmoonlightpay_video name=moonlight_pay \
payload_size={payload_size} fec_percentage={fec_percentage} min_required_fec_packets={min_required_fec_packets} !
appsink sync=false max-buffers=1 drop=true name=wolf_udp_sink
"""
```

This allowed me to reach stable `4K@60` (with the silly `150Mbps` settings in Moonlight); without the queue I was getting only 35 FPS.

## Next steps

I'd like to measure the **network impact** of using `sendmmsg` + pacing vs the previous implementation, but I'm not sure how to do that properly; any idea?

ABeltramo:
As for the memory leak, I can't reproduce that when running the latest image in Docker. I suppose something is wrong with my dev machine (Arch, Gstreamer `1.28.1`) but it's unrelated to these changes.

ABeltramo (approved):
I think this is in a good state now, I'll just wait to collect more feedback from users and possibly find a way to properly test the network usage with the new `sendmmsg` + pacing.  
Thanks @Baz00k for all your work on this!

ABeltramo:
I think there's something wrong with the Nvidia pipeline..
Here's a screenshot of setting 50Mbit/s when running on AMD using `vah265enc`
image
Perfectly flatlined, respecting the input bitrate.  
Here's the output when connecting to the same client but when using Nvidia (`nvh265enc`):
 
image

Flatlined but **it just uses 2MBit/s**.  Not sure if this is tied to the new parameters; in both cases I see that we actually end up passing to Gstreamer `bitrate=40108` but there's definitely something wrong going on with the Nvidia pipeline.. I just gave it a quick go so it could be something on my dev machine (again), is it just me?

Baz00k:
image
image

Just tested with Nvidia pipeline - the encoder seems to adjust the bitrate based on the screen content.
It scales with the requested bitrate and never exceeds it. Haven't done any deep dive yet.

ABeltramo:
Thanks for confirming, definitely looks like it's not using constant bitrate like `va` does. I'll play with the pipeline as soon as I get a bit of free time..

ABeltramo:
The Nvidia pipeline hasn't changed, so this behaviour was already present in `:stable`. Everything else looks good to me, time to merge this and see if we get any new issues or reports.    

Thanks to everyone involved in this PR! :hugs:

PR #378, opened 2026-03-29
gpu-drivers: add Kisak-Mesa PPA for latest Mesa drivers

JBailes:
## Summary
- Adds the [Kisak-Mesa PPA](https://launchpad.net/~kisak/+archive/ubuntu/kisak-mesa) to the GPU drivers container build
- GPU driver packages (`va-driver-all`, `intel-media-va-driver-non-free`, etc.) now resolve from the PPA's newer Mesa builds instead of the base image's default repositories
- `software-properties-common` is installed temporarily and purged after adding the PPA to keep the image lean

## Test plan
- [ ] Verify container builds successfully with the PPA
- [ ] Confirm Mesa driver versions in the built image come from Kisak-Mesa
- [ ] Test GPU acceleration functionality with the updated drivers

### DIFF
--- a/docker/gpu-drivers.Dockerfile
+++ b/docker/gpu-drivers.Dockerfile
@@ -12,12 +12,8 @@ ARG REQUIRED_PACKAGES="va-driver-all intel-media-va-driver-non-free \
                        libva-drm2 libva-x11-2 libvpl2"
 
 RUN apt-get update -y && \
-    apt-get install -y --no-install-recommends software-properties-common && \
-    add-apt-repository -y ppa:kisak/kisak-mesa && \
-    apt-get update -y && \
     apt-get install -y --no-install-recommends \
     $REQUIRED_PACKAGES && \
-    apt-get remove -y --purge software-properties-common && \
     rm -rf /var/lib/apt/lists/*
 
 # libmfx is not available in Ubuntu 25.04 so we are building from sources (see: https://github.com/games-on-whales/wolf/issues/221)
--- a/docker/gpu-drivers.Dockerfile
+++ b/docker/gpu-drivers.Dockerfile
@@ -12,8 +12,12 @@ ARG REQUIRED_PACKAGES="va-driver-all intel-media-va-driver-non-free \
                        libva-drm2 libva-x11-2 libvpl2"
 
 RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends software-properties-common && \
+    add-apt-repository -y ppa:kisak/kisak-mesa && \
+    apt-get update -y && \
     apt-get install -y --no-install-recommends \
     $REQUIRED_PACKAGES && \
+    apt-get remove -y --purge software-properties-common && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


ABeltramo:
Btw, we should move the Mesa PPA to the base GOW image (`base-app`), the driver needs to be the same in both Wolf and all the apps!

JBailes:
> Btw, we should move the Mesa PPA to the base GOW image (`base-app`), the driver needs to be the same in both Wolf and all the apps!

You got it. Let me get over this quirk with Cargo and Rust, then I'll add that to this PR!

JBailes:
The build failure isn't related to the Kisak-Mesa PPA changes. It's a pre-existing issue in `wolf.Dockerfile`.

`cargo install cargo-c@0.10.20` is failing because transitive dependencies (`cargo-credential-libsecret@0.5.5`, `cargo-util@0.2.27`, `crates-io@0.40.17`) have published new versions requiring rustc 1.92, but the image has rustc 1.91.1. Without `--locked`, cargo resolves the latest compatible versions of those deps rather than the ones that shipped with the release.

Fix: add `--locked` to pin deps to the versions from the `cargo-c@0.10.20` lockfile:

```dockerfile
cargo install cargo-c@0.10.20 --locked
```

The error message itself suggests this: "Try re-running `cargo install` with `--locked`"

ABeltramo:
Since https://github.com/games-on-whales/gow/pull/301 is merged we should retry building Wolf on top of that now

JBailes:
> Since [games-on-whales/gow#301](https://github.com/games-on-whales/gow/pull/301) is merged we should retry building Wolf on top of that now

Success, Abe!

docker/gpu-drivers.Dockerfile
```diff
@@ -12,8 +12,12 @@ ARG REQUIRED_PACKAGES="va-driver-all intel-media-va-driver-non-free \
                        libva-drm2 libva-x11-2 libvpl2"
 
 RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends software-properties-common && \
+    add-apt-repository -y ppa:kisak/kisak-mesa && \
+    apt-get update -y && \
     apt-get install -y --no-install-recommends \
     $REQUIRED_PACKAGES && \
+    apt-get remove -y --purge software-properties-common && \
     rm -rf /var/lib/apt/lists/*
```
ABeltramo, line 17: The changes here should be removed since it's already part of `base-app` 😉

ABeltramo (changes requested):

docker/gpu-drivers.Dockerfile
JBailes: Gotcha. But we are going to have to keep the cargo changes, hm...

JBailes: Fixed @ABeltramo , the cargo changes remain (Which are required to build Wolf), these are reverted.

JBailes:
Ready for merge @ABeltramo!

ABeltramo (approved):

PR #382, opened 2026-04-02
feat: add quickstart scripts for automated Wolf deployment

JBailes:
## Summary

- Adds the quickstart scripts (from [JBailes/wolf](https://github.com/JBailes/wolf/tree/main/quickstart)) into the official repo under `quickstart/`
- Updates all download URLs in the scripts to point to `games-on-whales/wolf` (stable branch) instead of `JBailes/wolf`
- Updates `quickstart.adoc` to prominently feature the automated setup with a clear `curl` one-liner, while preserving the existing manual Docker/Podman/Proxmox instructions under a "Manual setup" section

The quickstart scripts auto-detect the user's environment (Proxmox, LXC, Incus, Unraid, TrueNAS, NixOS, Docker, or Podman) and handle GPU detection, udev rules, NVIDIA driver volumes, and Wolf + Wolf Den deployment automatically. See the [quickstart README](https://github.com/games-on-whales/wolf/blob/aimee/quickstart-scripts/quickstart/README.md) for full documentation.

### Scripts included

| Script | Purpose |
|---|---|
| `wolf.sh` | Entry point: detects environment, downloads helpers, dispatches |
| `common.sh` | Shared helpers: GPU detection, compose generation, udev rules, NVIDIA volume |
| `configure.sh` | Container-side setup (used by Proxmox/LXC/Incus) |
| `docker.sh` | Docker Compose deployment |
| `podman.sh` | Podman Quadlet + systemd deployment |
| `proxmox.sh` | Proxmox LXC creation with GPU passthrough |
| `lxc.sh` | Standalone LXC container creation |
| `incus.sh` | Incus container creation |
| `unraid.sh` | Unraid deployment with boot persistence |
| `truenas.sh` | TrueNAS SCALE deployment with ZFS + midclt init scripts |
| `nixos.sh` | NixOS module generation (declarative) |

Supersedes #380.

## Test plan

- [ ] Verify `curl` command in quickstart.adoc downloads `wolf.sh` correctly from the `stable` branch once merged
- [ ] Test `wolf.sh` on Docker environment
- [ ] Test `wolf.sh` on Proxmox environment
- [ ] Verify quickstart.adoc renders correctly in Antora docs


docs/modules/user/pages/quickstart.adoc
```diff
@@ -2,9 +2,24 @@
 
 Wolf runs as a single container, it'll spin up and down additional containers on-demand.
 
-You can run the quickstart script directly with `curl -sSL https://raw.githubusercontent.com/jbailes/wolf/main/quickstart/wolf.sh | bash` or see the quickstart guide and script at https://github.com/JBailes/wolf/tree/main/quickstart for a fully automated script that will handle setting the environment up for you.
+== Automated setup (recommended)
```
ABeltramo: I wouldn't say this is recommended until we have some feedback from users actually running this. I'd put instead `beta, use at your own risk` or something along those lines

docs/modules/user/pages/quickstart.adoc
```diff
@@ -2,9 +2,24 @@
...
 
-== Docker
+The quickstart script auto-detects your environment (Proxmox, LXC, Incus, Unraid, TrueNAS, NixOS, Docker, or Podman) and sets up Wolf + Wolf Den automatically:
+
+[source,bash]
+....
+curl -fsSLO https://raw.githubusercontent.com/games-on-whales/wolf/stable/quickstart/wolf.sh
+chmod +x wolf.sh
+sudo ./wolf.sh
```
ABeltramo, line 18: Does it really need `sudo`? If it's required we should specify where and why that's required.

quickstart/configure.sh
```diff
@@ -0,0 +1,77 @@
...
+        chmod a+r /etc/apt/keyrings/docker.gpg
+
+        echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] \
+https://download.docker.com/linux/debian $(. /etc/os-release && echo "$VERSION_CODENAME") stable" \
+            > /etc/apt/sources.list.d/docker.list
+
+        apt-get update -qq
+        apt-get install -y --no-install-recommends \
+            docker-ce docker-ce-cli containerd.io docker-compose-plugin
+        info "Docker installed"
```
ABeltramo, line 26: We shouldn't do this, stuff like this gets out of date fast. We should error out possibly pointing to the official [Docker docs](https://docs.docker.com/get-started/).  

Also, how about users having Podman that don't want Docker installed?

quickstart/common.sh
```diff
@@ -0,0 +1,635 @@
...
+    docker compose -f /opt/wolf/docker-compose.yml pull
+    docker compose -f /opt/wolf/docker-compose.yml up -d
+
+    sleep 5
+    if docker compose -f /opt/wolf/docker-compose.yml ps --format '{{.Service}} {{.State}}' | grep -q "running"; then
+        info "Services are running"
+    else
+        warn "Some services may not be running yet. Check: docker compose -f /opt/wolf/docker-compose.yml ps"
+    fi
+}
```
ABeltramo: How about podman?

quickstart/common.sh
```diff
@@ -0,0 +1,635 @@
...
+
+[profiles.apps.runner]
+type = "docker"
+name = "WolfSteam"
+image = "ghcr.io/games-on-whales/steam:edge"
+mounts = ["/etc/wolf/steam:/home/retro:rw"]
+env = ["PROTON_LOG=1", "RUN_SWAY=true"]
+TOML
+}
+
```
ABeltramo, line 409: Why doing this? Wolf is perfectly capable of creating the default when it first starts. This isn't even a valid config file it's missing the encoders amongst other things.

quickstart/common.sh
```diff
@@ -0,0 +1,635 @@
...
+            ;;
+        AMD|Intel)
+            cat > "$compose_file" <<YAML
+services:
+  wolf:
+    image: ghcr.io/games-on-whales/wolf:stable
+    environment:
+      - WOLF_RENDER_NODE=${render_node}
+      - XDG_RUNTIME_DIR=/tmp/sockets
+      - WOLF_CFG_FILE=/etc/wolf/cfg/config.toml
```
ABeltramo: No need to rewrite something that is default already.

quickstart/nixos.sh
```diff
@@ -0,0 +1,319 @@
...
+  # Udev rules for Wolf virtual input devices
+  services.udev.extraRules = ''
+    KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput", TAG+="uaccess"
+    KERNEL=="uhid", GROUP="input", MODE="0660", TAG+="uaccess"
+    KERNEL=="hidraw*", ATTRS{name}=="Wolf PS5 (virtual) pad", GROUP="input", MODE="0660", ENV{ID_SEAT}="seat9"
+    SUBSYSTEMS=="input", ATTRS{name}=="Wolf X-Box One (virtual) pad", MODE="0660", ENV{ID_SEAT}="seat9"
+    SUBSYSTEMS=="input", ATTRS{name}=="Wolf PS5 (virtual) pad", MODE="0660", ENV{ID_SEAT}="seat9"
+    SUBSYSTEMS=="input", ATTRS{name}=="Wolf gamepad (virtual) motion sensors", MODE="0660", ENV{ID_SEAT}="seat9"
+    SUBSYSTEMS=="input", ATTRS{name}=="Wolf Nintendo (virtual) pad", MODE="0660", ENV{ID_SEAT}="seat9"
+  '';
```
ABeltramo: All these rules are repeated multiple times, we should have a single place where we set them and everything reads from one place. Otherwise this will go out of sync as soon as we need a change.

ABeltramo (changes requested):
Just took a quick look over this massive PR. I'm not convinced on the direction this is going.. It's already massive and with too many assumptions; supporting all possible combinations of user environments it's going to be painful.

quickstart/configure.sh
JBailes: This is only used when we generate a (LXC only perhaps) container (Or should be, you are having me look through all my assumptions now). I fixed the core issue you had, however.

docs/modules/user/pages/quickstart.adoc
JBailes: Yep, had this fixed awhile ago, but I now stress that it's beta ;)

docs/modules/user/pages/quickstart.adoc
JBailes: It depends on the system, truthfully, and user perms. This is really good feedback, I'm working on it now.

quickstart/common.sh
JBailes: Podman does have it's own scripts.

ABeltramo:
I was thinking about this script and I had an idea: instead of installing stuff and potentially mess up the user environment, why don't we output a file and an explanation on what to do with it?  

For example: in most cases we just need to output a `compose.yml` with the "right stuff" in it. For LXC it would be the config file. NixOS? Just output a flake.  

This way we only provide a suggested configuration:
- users can review and understand what's going on
- the script doesn't need `sudo` or any privilege
- we can't mess up anything because we don't install anything
- they can post the configuration in use when asking for help

This should simplify things massively, to the point where it mostly becomes like a good old wizard config 😅

JBailes:
> I was thinking about this script and I had an idea: instead of installing stuff and potentially mess up the user environment, why don't we output a file and an explanation on what to do with it?
> 
> For example: in most cases we just need to output a `compose.yml` with the "right stuff" in it. For LXC it would be the config file. NixOS? Just output a flake.
> 
> This way we only provide a suggested configuration:
> 
> * users can review and understand what's going on
> * the script doesn't need `sudo` or any privilege
> * we can't mess up anything because we don't install anything
> * they can post the configuration in use when asking for help
> 
> This should simplify things massively, to the point where it mostly becomes like a good old wizard config 😅

This is roughly my final point on where I want to go with the script. I view it as being 3 steps:

1) Script built
2) Script tested
3) Script split into OS-specific deployments, e.g. TRUENAS Community App, Unraid Community Plugin, Proxmox Community Script, etc.

JBailes:
> Just took a quick look over this massive PR. I'm not convinced on the direction this is going.. It's already massive and with too many assumptions; supporting all possible combinations of user environments it's going to be painful.



> > I was thinking about this script and I had an idea: instead of installing stuff and potentially mess up the user environment, why don't we output a file and an explanation on what to do with it?
> > For example: in most cases we just need to output a `compose.yml` with the "right stuff" in it. For LXC it would be the config file. NixOS? Just output a flake.
> > This way we only provide a suggested configuration:
> > 
> > * users can review and understand what's going on
> > * the script doesn't need `sudo` or any privilege
> > * we can't mess up anything because we don't install anything
> > * they can post the configuration in use when asking for help
> > 
> > This should simplify things massively, to the point where it mostly becomes like a good old wizard config 😅
> 
> This is roughly my final point on where I want to go with the script. I view it as being 3 steps:
> 
> 1. Script built
> 2. Script tested
> 3. Script split into OS-specific deployments, e.g. TRUENAS Community App, Unraid Community Plugin, Proxmox Community Script, etc.

To add more detail:

Bigger picture from here, with every wolf release, we'd release these automatically into the relevant community locations. Scripts would be there purely to generate the relevant community plugin/app/etc., which then makes it a "Go get our community app" installation.

ABeltramo:
I'm not sure I'm following, which community locations are you thinking about? What I was thinking was much simpler: a script that basically spits out the recommended config for the system in use. No more, no less than an assisted pick from one of the options on the quickstart guide.

JBailes:
> I'm not sure I'm following, which community locations are you thinking about? What I was thinking was much simpler: a script that basically spits out the recommended config for the system in use. No more, no less than an assisted pick from one of the options on the quickstart guide.

https://apps.truenas.com

Basically, my bigger goal is upon release, then it would create apps on various targets. I.e. there's been asks for unraid community plugins as another example. I think your idea is solid too, though.

PR #384, opened 2026-04-05
Replace existing controller on re-arrival instead of skipping

kmreisi:
## Summary

When a Moonlight client reconnects, it sends CONTROLLER_ARRIVAL for controllers that may already exist in the session (carried over via resume). Previously this was skipped with a TODO comment, leaving stale virtual devices in the container.

Now the old pad is unplugged (UnplugDeviceEvent) and removed from the joypads list before creating a fresh one, ensuring the container always has exactly one device per controller after reconnect.


kmreisi:
Closed in favour of https://github.com/games-on-whales/wolf/pull/385

PR #385, opened 2026-04-05
Add Nintendo Switch Pro Controller support

kmreisi:
## Summary

Adds full Nintendo Switch Pro Controller support, enabling:

- Gyroscope and accelerometer motion forwarding (requests motion events from Moonlight for Nintendo controller types, same as already done for PS5)
- Proper Pro Controller identity with Bluetooth bus, stable MAC-based UNIQ, and correct VID/PID so hid-nintendo binds and SDL recognizes the device
- Only hidraw and IMU evdev nodes are exposed to the container (the evdev joystick node is skipped to prevent SDL from creating a duplicate controller via both HIDAPI and evdev paths)
- Face button remapping (A↔B, X↔Y) via a `translate_buttons()` helper to compensate for SDL HIDAPI's Nintendo→Xbox remap on read
- Tests covering Switch metadata consistency, reconnect plug/unplug, and replayed Moonlight packet ranges

## Dependencies

Requires the inputtino Switch Pro UHID backend: games-on-whales/inputtino#37


src/moonlight-server/control/input_handler.cpp
```diff
@@ -17,6 +17,50 @@ using namespace wolf::core;
...
+    if (buttons & B)
+      out |= A;
+    if (buttons & X)
+      out |= Y;
+    if (buttons & Y)
+      out |= X;
+    return out;
+  }
+  return buttons;
+}
```
evertonstz: Is not the remap too opinionated to be in the server? We have an option to invert these in the moonlight clients.

kmreisi: I had some thoughts on this, at the end I decided that in this case it might make sense to have it in the server. Not sure on this :-D

When we do it like this, on the server, the "default" is to behave as written as on the controllers.

You can still use the client to "swap", and then it swaps and not restores the default....

I would be fine to change this :-)

evertonstz: Maybe the maintainers will have a better input on this xD

kmreisi: Lets see what the maintains say to this, I moved it btw to inputtino, but the outcome is the same, we map SDL buttons to the correct buttons on the controller:

https://github.com/games-on-whales/inputtino/pull/37/changes#diff-63e7815f1b2d8380301e2a5f77a53ba4dd248c2758bfa5a66d680f27f0c9c93aR492

:-)

kmreisi:
Superseded by #445: with wolf consuming inputtino's runtime joypad factory, Switch Pro (and now Joy-Con) support comes from inputtino directly, so the per-controller workarounds here are no longer needed. Closing in favour of #445.

PR #388, opened 2026-04-09
docs: fix error in steam path inside docker

ramadhanafif:
the current steam dir has moved from `~/.steam/debian-installation` to `~/.steam/steam`


ABeltramo:
I'm holding this off because we are also moving to Fedora so the Steam location might change again..

PR #390, opened 2026-04-12
fix: enable session_id_context for HTTPS server

BenSmith:
The `set_session_id_context` member of `Server<HTTPS>` defaults to false in rest.hpp, leaving the `after_bind() `override's call to `SSL_CTX_set_session_id_context()` as dead code. Wolf's HTTPS server therefore runs with no session ID context set on its `SSL_CTX`.

OpenSSL only enforces this on session-resumption-capable client cert handshakes. The first HTTPS request from a new client (the `/pair` challenge) goes through fine -- no resumption attempt. Every subsequent request from the same client (`/applist`, `/serverinfo`, `launch`, etc.) attempts session resumption, OpenSSL checks for a session ID context, finds none, and refuses with:

>  HTTPS error during request at <path> error code: 167772437 -  session id context uninitialized (SSL routines)

which surfaces on the client as `QNetworkReply::SslHandshakeFailedError` (or equivalent on other clients). The pairing succeeds but every subsequent operation fails.

Setting the default to `true` makes `after_bind()` do its job. Wolf's chosen `session_id_context` value (port + reversed address) works as a per-server unique tag.

Editing to add: I used AI tooling to help track this down.

### DIFF
--- a/src/moonlight-server/rest/rest.hpp
+++ b/src/moonlight-server/rest/rest.hpp
@@ -25,7 +25,7 @@ using HTTPS = asio::ssl::stream<asio::ip::tcp::socket>;
  *   while still be able to send back a reply in case of validation error
  */
 template <> class Server<HTTPS> : public ServerBase<HTTPS> {
-  bool set_session_id_context = false;
+  bool set_session_id_context = true;
 
 protected:
   asio::ssl::context context;
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,7 +22,8 @@ set(SRC_LIST
         testMoonlight.cpp
         testRTSP.cpp
         testWolfAPI.cpp
-        testBatchedSend.cpp)
+        testBatchedSend.cpp
+        testHTTPS.cpp)
 
 if (UNIX AND NOT APPLE)
     if (LINK_RUST_WAYLAND)
--- a/tests/testHTTPS.cpp
+++ b/tests/testHTTPS.cpp
@@ -1,0 +1,133 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include <atomic>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <curl/curl.h>
+#include <filesystem>
+#include <future>
+#include <rest/rest.hpp>
+#include <string>
+#include <thread>
+
+namespace fs = std::filesystem;
+
+namespace {
+
+struct TempDir {
+  fs::path path;
+  TempDir() {
+    auto base = fs::temp_directory_path() / "wolftests-https-XXXXXX";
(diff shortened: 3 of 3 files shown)


BenSmith:
This issue arose when I was building some Wolf support into my own project. Wolf is super cool! I have been poking at it off and on for a while now. I love the zero-copy implementation and merging a wayland compositor and a gstreamer plugin is some big brain stuff.

ABeltramo:
> which surfaces on the client as QNetworkReply::SslHandshakeFailedError (or equivalent on other clients). The pairing succeeds but every subsequent operation fails.

Is this something that was only failing in your custom code or did you find a Moonlight client that exibits that behaviour? I feel this would've been reported by now..

BenSmith:
> I feel this would've been reported by now..

I felt the same! 

I've merged the test PR into this one.

I was using the flatpak moonlight-qt client, and then I tried with a moonlight-qt client I compiled myself. And with an android moonlight client. None would maintain a connection before this change.

This code change does allow them all to connect, and it looks like it was there for this purpose, but what you're saying makes me think perhaps something in my configuration or pipeline is causing an initial disconnect that forces the client to attempt a resume, which then fails due to this setting?

So, conjecture, maybe you haven't received reports because I made things weird enough to trigger a TLS reconnect. I'll see if I can dig further into why my configuration is using the resume pathway. 

FWIW, [this is how I'm building Wolf](https://github.com/BenSmith/bootc-hypervisor/blob/7c62b50542f1cf503a3cbd14cefc5ac5932f4fa0/workloadctl/containers/wolf-game-streaming/Containerfile), and [this is its surrounding context](https://github.com/BenSmith/bootc-hypervisor/tree/7c62b50542f1cf503a3cbd14cefc5ac5932f4fa0/workloadctl/containers/wolf-game-streaming).

BenSmith:
Alright, well, it seems to work fine now. Which is kinda frustrating considering how much I put into shenanigans last weekend. My container version is pinned, github hashes pinned, etc. 

Regardless,  your official ghcr image worked, a version I built with your Dockerfile worked, my container build worked. With moonlight-qt from flatpak and source. I think this tls session resume thing could still be an issue, if it comes up again, but it's not showing up in my packet captures and I didn't capture any when making the report & PR.

Again thanks for the fantastic tool and ideas.

ABeltramo:
@BenSmith that's a shame, I get the frustration but I'm kind of glad that everything is working :sweat_smile:   
I'll keep an eye out for similar reports; if they ever surface I'll get back to this one!  

> Again thanks for the fantastic tool and ideas.

Thank you for taking the time to look into this and committing the code! Have fun with Wolf!

BenSmith:
Haha I get it. It's just time I could have been gaming 🙃

PR #391, opened 2026-04-12
test: HTTPS server must allow session resumption with client cert

BenSmith:
Adds an integration test that spins up Wolf's Server<HTTPS> on an ephemeral port and issues two consecutive TLS handshakes with a client certificate over a reusable libcurl handle. The second handshake exercises OpenSSL's session resumption path, which aborts with "session id context uninitialized" unless the server has called `SSL_CTX_set_session_id_context()` on its `SSL_CTX`.

On stable this test currently fails: the [fix lives in a separate PR](https://github.com/games-on-whales/wolf/pull/390).

I did not see other similar integration tests and I am not trying to set precedent for them with this PR, feel free to ignore this if you do not want it.

Edited to add: This is AI-generated code.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -22,7 +22,8 @@ set(SRC_LIST
         testMoonlight.cpp
         testRTSP.cpp
         testWolfAPI.cpp
-        testBatchedSend.cpp)
+        testBatchedSend.cpp
+        testHTTPS.cpp)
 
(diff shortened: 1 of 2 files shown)


ABeltramo:
Thanks for opening this! Can you merge this test with your other PR so that I can review them together?

BenSmith:
Will do, I didn't want to presume to impose an integration testing regime with the code change

PR #393, opened 2026-04-18
Fix session containers to run as retro

JBailes:
## Summary
- set `UNAME=retro` for Wolf-launched session containers alongside the existing `PUID` and `PGID` overrides
- add a regression test around `sessions::start_runner` so session container env construction keeps the non-root runtime user

## Root Cause
Wolf already overrides the numeric runtime IDs for session containers, but it does not override `UNAME`. GoW images such as `wolf-ui` default that variable to `root`, and the base entrypoint only switches to the non-root path when `UNAME != root`. That leaves session containers running as root even when Wolf is trying to launch them with the paired client's runtime UID/GID.

## Validation
- `git diff --check`
- added a focused `start_runner` regression test covering the session env

### DIFF
--- a/src/moonlight-server/sessions/common.cpp
+++ b/src/moonlight-server/sessions/common.cpp
@@ -65,6 +65,7 @@ void start_runner(std::shared_ptr<events::Runner> runner,
 
   full_env.set("PUID", std::to_string(args->client_settings->run_uid));
   full_env.set("PGID", std::to_string(args->client_settings->run_gid));
+  full_env.set("UNAME", "retro");
 
   // Add fake-udev and udev mounts
   mounted_paths.push_back(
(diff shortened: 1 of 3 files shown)


JBailes:
@ABeltramo Be gentle, this is the first C++ I've written in many years. I ran into this as part of debugging the steam image and decided to finally fix it.

ABeltramo:
I don't think we should hardcode `UNAME=retro` in Wolf. Our images already bake `retro` and if the user wants to run different images (or indeed a different user) having an hardcoded env variable like that could cause issues. 

```bash
docker run --rm -it --entrypoint /bin/bash ghcr.io/games-on-whales/steam:edge               
root@d3c4aa7e2831:~# echo $UNAME
retro
```

PR #396, opened 2026-04-20
Add party mode split rendering

JBailes:
## What changed
Adds 2-seat party-mode split rendering to Wolf, layered on top of #419's zero-copy / gst-interpipe context-sharing work.

- **Rebased onto #419** (`fix/amd-dma-buf-negotiation`) — this PR now targets #419, not `stable`, because the GPU-aware split compositor depends on #419's gst-interpipe `GST_QUERY_CONTEXT` forwarding (so VA/CUDA elements share a context across the interpipe boundary) and its generic-DMABuf runtime negotiation.
- **Party-mode session management + API** for join, promote, and teardown; secondary "hidden seat" sessions; local-gamepad join path.
- **GPU-aware split compositor.** The split is composited entirely in GPU memory, selected per detected vendor so solo and party sessions stay zero-copy:
  - VA/QSV: `vapostproc` + `vacompositor` (VAMemory)
  - NVIDIA: `cudaconvertscale` + `cudacompositor` (CUDAMemory)
  - software fallback: `videoconvertscale` + `compositor`
  The old unconditional system-memory `videoconvertscale` could not ingest CUDAMemory/DMABuf at all, so it would have broken zero-copy streaming for every GPU session; the per-vendor sources fix that.
- **Config bumped to v9** (migration `version <= 8`) so existing v8 (10-bit) installs are re-migrated and pick up the party pipeline. `config.v9.toml` carries the software party source plus `default_source_va` / `default_source_cuda`; `configTOML` selects by vendor.
- Restart-runner-in-place so the primary app relaunches at split render size without dropping the Moonlight session; render size decoupled from `display_mode` via `render_display_mode`.
- Hardened session-id parsing across all session endpoints (`parse_session_id`), preserve producer caps features on resize, dedup the two-seat geometry.

## Verification
- `cmake` build of wolf (full Docker image) green against #419's pinned gst-interpipe.
- VA (`vacompositor`) and CUDA (`cudacompositor`) split-compositor chains validated standalone with `gst-launch` on real hardware — both negotiate the split entirely in GPU memory (no host round-trip), reach PLAYING, no caps errors.
- Real wolf binary boots, migrates config to v9, and selects the correct GPU party source per vendor: "zero copy pipeline on Nvidia" (RTX 5080, renderD128 -> CUDA source) and "zero copy pipeline on Intel" (UHD 770, renderD129 -> VA source).
- Pending: end-to-end Moonlight client stream with a live party join (decode/visual check across vendors).

## Depends on
- #419 (`fix/amd-dma-buf-negotiation`) — base branch.
- `games-on-whales/wolf-ui#12`, `games-on-whales/gst-wayland-display#27` (companion changes).


JBailes:
Promoting this to ready for review.

JBailes:
PR image handling update:

- Added branch `pr-396-fedora-image` for the Fedora validation image work.
- Changed the release Docker workflow on that branch so generic image publication only runs on `master`/`stable`, preventing PR branch pushes from writing broad tags such as `:edge`, `:fedora`, or `:ubuntu`.
- Added a PR-image workflow that tags images only as `pr-{PR_ID}-{OS}`; the branch form `pr-{PR_ID}-{OS}-image` is parsed automatically.
- For PR #396, `pr-396-fedora-image` resolves to `pr-396-fedora`; the Ubuntu matrix job is skipped and no Ubuntu image is built.
- The workflow builds the isolated Fedora chain under PR-specific tags: `gpu-drivers:pr-396-fedora`, `gstreamer:pr-396-fedora`, and `wolf:pr-396-fedora`.
- The first Fedora run failed because `intel-media-driver` no longer resolves on Fedora 43. I updated the Fedora GPU-driver Dockerfile to use `dnf --skip-unavailable` for that optional package and restarted the PR-image run.

Current Fedora-only run: https://github.com/games-on-whales/wolf/actions/runs/25916062709

JBailes:
@ABeltramo The latest change to this changes how PRs behave. It's pissed me off for awhile PRs can come in and kool aid man other tags. Instead, PRs now will have the specific image tag of pr-{PR_ID}-{OS} and won't be able to have other tags unless manually set. No more blowing away :edge randomly by a PR anymore.

ABeltramo:
Makes sense, I always used the `sha` tag tbh. Nobody should use `:edge`, if anything we should probably remove that tag..

JBailes:
> Makes sense, I always used the `sha` tag tbh. Nobody should use `:edge`, if anything we should probably remove that tag..

I would agree, this should be changed, but in a different PR. This is kind of "kitchen-sinky" already.

JBailes:
@ABeltramo Refactored this so it is only the party mode PR and ready to merge.

JBailes:
I am rewriting this in the context of #419 , as it exposes a lot of stuff that directly impact party mode (And will end up with a substantially better party mode).

Wait until after the next push to the PR to do anything with this, and it will have a hard dependency on #419 being merged.

ABeltramo:
Since we are bumping the config, we should remove `PROTON_LOG=1` from the default for Steam, see #434

Issue #394, opened 2026-04-19
Wolf ui no longer connecting to wayland/X11


eco9898:
I set up the container using the [quickstart](https://games-on-whales.github.io/wolf/stable/user/quickstart.html) and have been using it happily until today the wolf UI no longer loads. I have moved my wolf volume to a new disk, but no other changes. Was I meant to change something in the config files with the new path.

The test app still works and nvidia-smi shows wolf as an active process.

```yaml
services:
  wolf:
    deploy:
      resources:
        reservations:
          devices:
            - capabilities:
                - gpu
              count: all
              driver: nvidia
    device_cgroup_rules:
      - c 13:* rmw
    devices:
      - /dev/dri
      - /dev/uinput
      - /dev/uhid
    environment:
      - NVIDIA_DRIVER_CAPABILITIES=all
      - NVIDIA_VISIBLE_DEVICES=all
      - WOLF_PULSE_CONTAINER_TIMEOUT_MS=30000
      - WOLF_STOP_CONTAINER_ON_EXIT=false
    image: ghcr.io/games-on-whales/wolf:stable
    network_mode: host
    restart: unless-stopped
    volumes:
      - /new/disk/wolf/data:/etc/wolf
      - /var/run/docker.sock:/var/run/docker.sock:rw
      - /dev/:/dev/:rw
      - /run/udev:/run/udev:rw
version: '3'
```

Container logs for wolf-ui
```
[2026-04-19 04:36:43] 
[2026-04-19 04:36:43] [ /etc/cont-init.d/10-setup_user.sh: executing... ]
[2026-04-19 04:36:43] **** Configure default user ****
[2026-04-19 04:36:43] Container running as root. Nothing to do.
[2026-04-19 04:36:43] DONE
[2026-04-19 04:36:43] 
[2026-04-19 04:36:43] [ /etc/cont-init.d/15-setup_devices.sh: executing... ]
[2026-04-19 04:36:43] **** Configure devices ****
[2026-04-19 04:36:43] Exec device groups
[2026-04-19 04:36:43] Path '/dev/input/event*' is not present.
[2026-04-19 04:36:43] Adding user 'root' to groups: root
[2026-04-19 04:36:43] DONE
[2026-04-19 04:36:43] 
[2026-04-19 04:36:43] [ /etc/cont-init.d/30-nvidia.sh: executing... ]
[2026-04-19 04:36:43] Nvidia driver detected, assuming it's using the nvidia driver volume
[...]

PR #397, opened 2026-04-23
Fix Wayland startup race and AMD DMA-BUF negotiation

JBailes:
## Summary
- mount the full XDG runtime dir into runner containers instead of bind-mounting individual Wayland/Pulse sockets
- wait for the compositor socket to exist before starting stream runners
- fix AMD zero-copy DMA-BUF negotiation by intersecting `waylanddisplaysrc` source caps with `vapostproc` sink caps before enabling the pipeline
- skip fork-PR auto-push behavior in the `clang-format` workflow so CI validates formatting without failing on detached HEAD

## Why
Wolf UI startup on Proxmox LXC exposed two separate issues:
- the runner could start before the compositor socket existed, which caused the app container to fail with `Can't connect to a Wayland display`
- the AMD zero-copy path could advertise DMA-BUF formats that `waylanddisplaysrc` could not actually provide to `vapostproc`, which caused the producer pipeline to fail negotiation during stream startup

Mounting the full runtime dir and waiting for the Wayland socket removes the startup race. Intersecting the producer and consumer DMA-BUF caps fixes the real zero-copy negotiation bug instead of forcing Wolf UI onto legacy `video/x-raw` caps.

## Validation
- reproduced the failure against Wolf UI on Proxmox LXC + RX 7900 XTX
- verified that the runtime-dir/socket changes address the Wayland display startup failure
- verified that the DMA-BUF caps intersection keeps zero-copy enabled only when `waylanddisplaysrc` and `vapostproc` share compatible formats
- ran `git diff --check`


JBailes:
Fix for https://github.com/games-on-whales/wolf/issues/394#issuecomment-4299715156

src/moonlight-server/sessions/common.cpp
```diff
@@ -20,17 +20,16 @@ void start_runner(std::shared_ptr<events::Runner> runner,
   full_env.set("XDG_RUNTIME_DIR", args->xdg_runtime_dir);
   full_env.set("WOLF_SESSION_ID", args->session_id);
 
+  // Mount the whole runtime directory into the runner container. This avoids
+  // bind-mount races on per-socket paths when the Wayland socket is created
+  // shortly after the compositor announces its display name.
+  mounted_paths.push_back({args->host->host_xdg_runtime_dir, args->xdg_runtime_dir});
+
```
ABeltramo: This would break isolation though, every app container can access all the wayland sockets

tests/main.cpp
```diff
@@ -6,14 +6,16 @@
...
 #include <helpers/utils.hpp>
 #include <streaming/streaming.hpp>
+#include <cstdlib>
 
 int main(int argc, char *argv[]) {
   logs::init(logs::parse_level(utils::get_env("WOLF_LOG_LEVEL", "TRACE")));
   streaming::init();
   control::init();       // Need to initialise enet once
   state::docker::init(); // Need to initialise libcurl once
+  setenv("WOLF_SKIP_WAYLAND_SOCKET_WAIT", "TRUE", 1);
```
ABeltramo: Usually put this stuff in the Dockerfile instead. The result is the same, but it's just more visible to have all the env variables in the same place (with the defaults too) instead of having to search in the codebase..

src/moonlight-server/sessions/common.cpp
JBailes: Hmm. Good point. I'm going to have to think more on how to solve this.

JBailes: Did a quirk and dirty initial fix, but my "quick and dirty" may be all the fix that is needed in this case.

tests/main.cpp
JBailes: This was a test issue, I think I've rewritten it in a way that keeps what you asked for but keeps it strictly to testing. I don't like putting test-only stuff in dockerfiles.

ABeltramo: Sorry I missed that this was the `main.cpp` for the tests; makes perfect sense to keep it there, discard all I've said 😅

JBailes: I mean, I can revert the change, but it is technically a bit better now.

JBailes: @ABeltramo Okay, I started pulling at the test thread here and couldn't stop. Take a look at the changes, this should be an improvement and, more importantly, replicable across many systems.

src/moonlight-server/sessions/common.hpp
```diff
@@ -1,6 +1,48 @@
...
+#include <thread>
+
+#include <helpers/logger.hpp>
 
 namespace wolf::core::sessions {
 constexpr std::string_view VIRTUAL_SINK_PREFIX = "virtual_sink_";
+
+inline bool wait_for_wayland_socket(std::string_view runtime_dir,
+                                    const std::string &socket_name,
+                                    std::chrono::milliseconds timeout = std::chrono::seconds(5)) {
```
ABeltramo: I think this should be an env variable. I imagine, like the Pulse socket, the time it takes for this to be actually ready is dependent on how fast the HW is..

JBailes: I'd be down. You got a number in mind as default @ABeltramo ?

JBailes: Now configurable, I set it as default 5s waiting on your feedback Abe.

ABeltramo: Sounds good to me, I just don't see the commit :sweat_smile:

JBailes: Derp, good catch, I forgot the push.

eco9898:
Just bumping this but I understand if everyone involved is still busy.

JBailes:
This should be good to go, just waiting on Abe's okay.

JBailes:
@ABeltramo Hoping to get this in, updated it to fix the merge conflicts.

ABeltramo:
I've tested this out locally with the latest `gst-wayland-display:stable` which results in 
```
09:00:40.210235086 DEBUG | Required DMA formats for vapostproc: ["NV12:0x0200000000082305", "YV12:0x0200000000082305", "YU12:0x0200000000082305", "P010:0x0200000000082305", "P012:0x0200000000082305", "R8  :0x0200000000082305", "YUYV", "YU24:0x0200000000082305", "AB24:0x0200000000082305", "AR24:0x0200000000082305", "XB24:0x0200000000082305", "XR24:0x0200000000082305", "BA24:0x0200000000082305", "AR30:0x0200000000082305"]
09:00:40.210251416 DEBUG | Available DMA formats for waylanddisplaysrc: []
09:00:40.210304536 WARN  | Unable to find any compatible DMA formats between waylanddisplaysrc and vapostproc, disabling zero copy pipeline.
```
Looking at the code in `gstreamer.hpp` when we iterate over the `GstCaps` we look for
```c++
gst_structure_get_list(structure, "drm-format", &formats);
```
but our custom compositor doesn't expose that 
```
gst-inspect-1.0 waylanddisplaysrc
[...]
Pad Templates:
  SRC template: 'src'
    Availability: Always
    Capabilities:
      video/x-raw(memory:DMABuf)
                 format: DMA_DRM (string)
                  width: [ 1, 2147483646 ] (GstIntRange)
                 height: [ 1, 2147483646 ] (GstIntRange)
              framerate: [ 1/1, 2147483646/1 ] (GstFractionRange)
      video/x-raw
                 format: RGBx (string)
                  width: [ 1, 2147483646 ] (GstIntRange)
                 height: [ 1, 2147483646 ] (GstIntRange)
              framerate: [ 1/1, 2147483646/1 ] (GstFractionRange)
      video/x-raw(memory:CUDAMemory)
                 format: { (string)BGRA, (string)RGBA } (GstValueList)
                  width: [ 1, 2147483646 ] (GstIntRange)
                 height: [ 1, 2147483646 ] (GstIntRange)
              framerate: [ 1/1, 2147483646/1 ] (GstFractionRange)
```
like `vapostproc` does
```
      video/x-raw(memory:DMABuf)
                  width: [ 1, 65536 ] (GstIntRange)
                 height: [ 1, 65536 ] (GstIntRange)
                 format: DMA_DRM (string)
             drm-format: { (string)NV12:0x0200000000082305, (string)YV12:0x0200000000082305, (string)YU12:0x0200000000082305, (string)P010:0x0200000000082305, (string)P012:0x0200000000082305, (string)"R8\ \ :0x0200000000082305", (string)YUYV, (string)YU24:0x0200000000082305, (string)AB24:0x0200000000082305, (string)AR24:0x0200000000082305, (string)XB24:0x0200000000082305, (string)XR24:0x0200000000082305, (string)BA24:0x0200000000082305, (string)AR30:0x0200000000082305 } (GstValueList)
```
so the added manual intersection wouldn't work, unless I'm missing something..?  

----

Now that I think about this, maybe this needs a full re-think..  

The way it works in `gst-wayland-comp` is that basically we publish the actual DMA formats **at runtime** based on the `render_node` param. This is supported in Gstreamer, and it's actually encouraged since you can't know the actual formats until you actually pick a GPU. Now what would happen when you have 2 `vaapi` capable GPUs? `vapostproc` most likely would return the formats of `renderD128` but that might not be the one that the user wants to pick and so it'll cause issues down the line later on.  

In an ideal world Gstreamer would do all the caps intersection at runtime without having us to manually fiddle with it, the problem is that `gstinterpipe` doesn't support that [gst-interpipe#111](https://github.com/RidgeRun/gst-interpipe/issues/111), [gst-interpipe#165](https://github.com/RidgeRun/gst-interpipe/pull/165). So maybe it's time that we try again with the power of LLMs to fix this properly once and for all?

src/moonlight-server/sessions/common.hpp
```diff
@@ -1,6 +1,71 @@
...
+#include <thread>
+
+#include <helpers/logger.hpp>
 
 namespace wolf::core::sessions {
 constexpr std::string_view VIRTUAL_SINK_PREFIX = "virtual_sink_";
+constexpr std::chrono::milliseconds DEFAULT_WAYLAND_SOCKET_WAIT_TIMEOUT = std::chrono::seconds(5);
+
+inline std::chrono::milliseconds get_wayland_socket_wait_timeout() {
+  if (const char *timeout_ms = std::getenv("WOLF_WAYLAND_SOCKET_WAIT_TIMEOUT_MS")) {
```
ABeltramo, line 19: Same as the other PR, let's use `utils::get_env` and add it to the docs and Dockerfile

ABeltramo:
I propose to split out this PR, we can merge the rest of the fixes and work on fixing the DMA caps on a separate PR..

JBailes:
> I propose to split out this PR, we can merge the rest of the fixes and work on fixing the DMA caps on a separate PR..

Works for me. I'll have something this evening, or you can go ahead and split right now.

ABeltramo:
No rush, I'd like to get a second set on eyes on the additional Boost crap that I'm fixing :sweat_smile:

JBailes:
Split this as discussed:

- **#418** — Wolf UI Wayland startup race (runtime-dir mount + socket wait), CI fixes, and the Boost build fixes. This is the mergeable half.
- **#419** — the AMD DMA-BUF caps intersection on its own. As noted above, `waylanddisplaysrc` advertises `DMA_DRM` without a `drm-format` list so the static-caps intersection currently disables zero copy — that PR is the starting point for re-thinking the negotiation properly.

Both branches live directly on `games-on-whales/wolf`. Suggest closing this one in favour of #418 + #419.

PR #408, opened 2026-05-05
Harden NVIDIA device and PulseAudio startup handling

JBailes:
## Summary
- skip linked NVIDIA device paths unless they are character devices before passing them to Docker app containers
- set HOME=/home/retro when Wolf launches the PulseAudio helper container
- update the NVIDIA linked-device test to reflect optional /dev/nvidia-uvm-tools availability

## Testing
- git diff --check
- attempted CMake configure with `cmake -S /home/virant/gow/.aimee-wolf-019df7dc -B /tmp/wolf-pr-build -DBUILD_TESTING=ON`; stopped after dependency FetchContent remained stuck in Boost download for several minutes

### DIFF
--- a/src/gst-video-context/gst-video-context/stub/cuda.h
+++ b/src/gst-video-context/gst-video-context/stub/cuda.h
@@ -39,6 +39,10 @@ typedef guint64 CUtexObject;
 typedef guintptr CUdeviceptr;
 typedef gint CUdevice;
 
+typedef struct CUuuid_st {                                /**< CUDA definition of UUID */
+  char bytes[16];
+} CUuuid;
+
 typedef enum {
   CUDA_SUCCESS = 0,
   CUDA_ERROR_NO_DEVICE = 100,
--- a/src/gst-video-context/gst-video-context/stub/cuda.h
+++ b/src/gst-video-context/gst-video-context/stub/cuda.h
@@ -39,7 +39,7 @@ typedef guint64 CUtexObject;
 typedef guintptr CUdeviceptr;
 typedef gint CUdevice;
 
-typedef struct CUuuid_st {                                /**< CUDA definition of UUID */
+typedef struct CUuuid_st { /**< CUDA definition of UUID */
   char bytes[16];
 } CUuuid;
 
--- a/src/moonlight-server/platforms/hw_linux.cpp
+++ b/src/moonlight-server/platforms/hw_linux.cpp
@@ -12,6 +12,7 @@
 #include <net/ethernet.h>
 #include <netinet/in.h>
 #include <optional>
+#include <system_error>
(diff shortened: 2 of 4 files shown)


ABeltramo:
> skip linked NVIDIA device paths unless they are character devices before passing them to Docker app containers

Why? What's the reason behind this change?

JBailes:
TL;DRL: For us users of LXC containers, NVIDIA device files that are normally character devices can exist as regular placeholder files instead. Specifically, /dev/nvidia-uvm-tools inside the LXC container was a regular file rather than a character device.

When Wolf enumerates NVIDIA devices and passes them to Docker for app containers, Docker performs a device-type check. If a "device" path is actually a plain file, Docker throws an error like "error gathering device information while adding custom device" and fails to start the container which was causing the Moonlight black screen / crash for me yesterday on nVidia.

The fix in PR #408 adds a guard: skip any linked NVIDIA device path that isn't actually a character device before handing the list to Docker. This prevents  the failure mode where LXC placeholder files for devices that aren't fully mapped into the container that cause Docker startup to abort.

ABeltramo:
Wouldn't things fail later on (for example when starting up games from Steam) if the Nvidia additional devices aren't present/working? Should we fail instead of spinning up a container that will end up not working properly?

JBailes:
In this case, what is being hardened against isn't actually needed to run the containers.

However, an alternative route could exist where we fail with a message the user can understand about what the actual problem is, instead of something that is largely incomprehensible.

ABeltramo:
> an alternative route could exist where we fail with a message the user can understand about what the actual problem is, instead of something that is largely incomprehensible.

I think this would be the proper solution here instead of starting a container that will then not work properly anyway. Otherwise, we would get more reports of weird issues that we have no visibility over IIUC

JBailes:
That's fair, but keep in mind that this causes the container to work properly. I have to run, but the tl;dr here is that we don't actually need these components for some reason. I don't know the deeper technical reason completely yet (This is a part I am still digging into technically), it's something I am working on.

JBailes:
@ABeltramo Okay, it took some time for me to understand the full profile. Uvm-tools does *not* matter to rendering, gaming, or anything else that we care about. All it is is a profiling/debugging interface. It allows people to use tools like NSight...but we don't care about NSight or other tooling.

For any of the other NVIDIA binaries, I wouldn't mind throwing a warning....but this change doesn't address that case, only specifically uvm-tools. The rest of the changes are just defensive programming changes, I'd be happy to explain or remove any defensive programming change you want out of the commit.

JBailes:
@ABeltramo Updated it so it does what we both want. Tell me what you think about this now!

ABeltramo (approved):

PR #415, opened 2026-05-31
Make default run UID/GID configurable, fix UID following GID on update

JBailes:
## Problem

Apps run as `1000:1000` by default, which bakes in a UID/GID convention that
isn't universal. On **Unraid** the `nobody` user is `99:100`, so the defaults
never matched the host and every newly paired client had to be corrected by
hand.

That manual correction was also broken. In `endpoint_UpdateClientSettings`,
the `run_uid` field was assigned from `new_settings.run_gid`:

```cpp
.run_uid = new_settings.run_gid.value_or(current_settings.run_uid),
.run_gid = new_settings.run_gid.value_or(current_settings.run_gid),
```

So whatever GID you submitted was copied into the UID too — you could only ever
end up with `99:99` or `100:100`, never `99:100`. Submitting only a new UID
(no GID) silently dropped the change, because the `value_or` fell through to the
current value. On Unraid this left Steam and other apps running with a UID
outside the expected `1000–60000` range and with file ownership that didn't line
up with the host, which broke streaming.

## Changes

- **Configurable defaults via env vars.** `ClientSettings::run_uid` / `run_gid`
  now read `WOLF_DEFAULT_RUN_UID` / `WOLF_DEFAULT_RUN_GID` (falling back to
  `1000`). Both new-client paths (`endpoint_Pair`-time session creation and the
  pairing event handler) default-construct `ClientSettings`, so they inherit the
  override. Persisted clients keep their saved values, so this only affects
  *new* clients. Malformed/empty values fall back to `1000`.
- **Fix the UID-follows-GID bug.** `endpoint_UpdateClientSettings` now reads
  `new_settings.run_uid` for the `run_uid` field, so UID and GID can be set
  independently.
- **Docs.** Document the two new environment variables in the quickstart.

## Why env vars

This is what lets the Unraid plugin (and any other turnkey deployment) ship the
right defaults: the plugin controls Wolf's container environment, so it can set
`WOLF_DEFAULT_RUN_UID=99` / `WOLF_DEFAULT_RUN_GID=100` and every new client is
correct out of the box, with no per-client editing.

## Testing

The new `env_default_run_id` helper was compiled standalone with
`-std=c++17 -Wall -Wextra` and verified for the unset / valid / malformed cases.
I was not able to run a full Wolf build locally; relying on CI for the
integration build.

### DIFF
--- a/src/moonlight-server/state/serialised_config.hpp
+++ b/src/moonlight-server/state/serialised_config.hpp
@@ -1,6 +1,8 @@
 #pragma once
 
+#include <cstdlib>
 #include <rfl.hpp>
+#include <string>
 
 namespace wolf::config {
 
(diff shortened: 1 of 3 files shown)


src/moonlight-server/state/serialised_config.hpp
```diff
@@ -11,9 +13,28 @@ enum class ControllerType {
...
+inline uint env_default_run_id(const char *env_var, uint fallback) {
+  if (const char *value = std::getenv(env_var)) {
+    try {
+      return static_cast<uint>(std::stoul(value));
+    } catch (...) {
+      // Ignore malformed values and use the fallback instead.
+    }
+  }
+  return fallback;
+}
```
ABeltramo: There's already a `get_env` utility that supports a fallback too. The cast to int shouldn't be needed since env vars are passed as strings IIRC

ABeltramo (changes requested):

JBailes:
Thanks for the review! Addressed in 113adfe:

- Now using the existing `utils::get_env("WOLF_DEFAULT_RUN_UID", "1000")` with its string fallback instead of a bespoke wrapper, matching the `std::stoi(utils::get_env(...))` pattern already in `data-structures.hpp`.
- Dropped the cast.

On CI: the `build-docs` failure is the final repository-dispatch step (`POST .../dispatches`) returning `401 Bad credentials` because the token isn't available to fork PRs — it's unrelated to this change. The Antora build itself runs before that step. Happy to adjust the doc note if you'd prefer it elsewhere.

JBailes:
Superseded by an in-repo branch (fork PRs can't access CI secrets, so build-docs always 401s). Continuing in https://github.com/games-on-whales/wolf/pull/416.

PR #416, opened 2026-05-31
Make default run UID/GID configurable, fix UID following GID on update

JBailes:
## Problem

Apps run as `1000:1000` by default, which bakes in a UID/GID convention that
isn't universal. On **Unraid** the `nobody` user is `99:100`, so the defaults
never matched the host and every newly paired client had to be corrected by
hand.

That manual correction was also broken. In `endpoint_UpdateClientSettings`,
the `run_uid` field was assigned from `new_settings.run_gid`:

```cpp
.run_uid = new_settings.run_gid.value_or(current_settings.run_uid),
.run_gid = new_settings.run_gid.value_or(current_settings.run_gid),
```

So whatever GID you submitted was copied into the UID too — you could only ever
end up with `99:99` or `100:100`, never `99:100`. Submitting only a new UID
(no GID) silently dropped the change, because the `value_or` fell through to the
current value. On Unraid this left Steam and other apps running with a UID
outside the expected `1000–60000` range and with file ownership that didn't line
up with the host, which broke streaming.

## Changes

- **Configurable defaults via env vars.** `ClientSettings::run_uid` / `run_gid`
  now read `WOLF_DEFAULT_RUN_UID` / `WOLF_DEFAULT_RUN_GID` via the existing
  `utils::get_env` (falling back to `1000`). Both new-client paths default-construct
  `ClientSettings`, so they inherit the override. Persisted clients keep their
  saved values, so this only affects *new* clients.
- **Fix the UID-follows-GID bug.** `endpoint_UpdateClientSettings` now reads
  `new_settings.run_uid` for the `run_uid` field, so UID and GID can be set
  independently.
- **Docs.** Document the two new environment variables in the quickstart.

## Why env vars

This is what lets the Unraid plugin (and any other turnkey deployment) ship the
right defaults: the plugin controls Wolf's container environment, so it can set
`WOLF_DEFAULT_RUN_UID=99` / `WOLF_DEFAULT_RUN_GID=100` and every new client is
correct out of the box, with no per-client editing.


ABeltramo (changes requested):
Looks good! Just a couple more changes:
- Add the new env variables in the docs under the list in the configuration page
- I also tend to add them in the Dockerfile, mainly for visibility but it's also good that everyone will basically use the env path instead of relying on the hardcoded value 

Thanks!!

JBailes:
@ABeltramo done

ABeltramo (approved):
Thanks, LGTM!

PR #418, opened 2026-06-01
Fix Wolf UI Wayland startup race (runtime-dir mount + socket wait)

JBailes:
## Summary
Split out of #397 — this is the mergeable half. The AMD DMA-BUF caps negotiation is moved to a separate PR per review discussion.

- mount the full XDG runtime dir into runner containers instead of bind-mounting individual Wayland/Pulse sockets
- wait for the compositor socket to exist before starting stream runners (with a configurable timeout)
- skip fork-PR auto-push behavior in the `clang-format` workflow so CI validates formatting without failing on detached HEAD
- Boost build fixes (simplified Boost loading in CMake + `libboost-json-dev`, Boost 1.91 `json::object` initializer)

## Why
Wolf UI startup on Proxmox LXC could start the runner before the compositor socket existed, causing the app container to fail with `Can't connect to a Wayland display`. Mounting the full runtime dir and waiting for the Wayland socket removes that startup race.

## Validation
- reproduced the failure against Wolf UI on Proxmox LXC + RX 7900 XTX
- verified the runtime-dir/socket changes address the Wayland display startup failure
- ran `git diff --check`


JBailes:
@ABeltramo, had a LLM summarize what I found:

1. The newer-Boost path the change exists for is untested in CI. Removing EXACT only does anything when system Boost is >= 1.87. The native CI builds on ubuntu-24.04 (Boost 1.83) and the Docker base is ghcr.io/games-on-whales/gstreamer:1.26.7 (Ubuntu, Boost 1.74/1.83) — both below the 1.87 floor. So find_package(Boost 1.87 …) always fails there and the whole of Boost 1.87 (json included) is FetchContent-compiled. The 1.91 behavior that motivated the fix (ABeltramo's Arch box) is dev-only and never exercised by CI — a future incompatibility on newer Boost won't be caught.
2. libboost-json-dev (added to the Dockerfile and linux-build-test.yml) is effectively inert. Following from #1: because system Boost < 1.87 in every CI/Docker build, Boost_FOUND is false and Boost is built from source regardless — so none of the libboost-*-dev apt packages actually feed the Boost link. Adding libboost-json-dev extends an already-dead list. It's only meaningful on distros shipping Boost ≥ 1.87. Not harmful, but misleading, and it implies a system-Boost build that isn't happening.
3. BOOST_VERSION now means two different things. After dropping EXACT, 1.87.0 is simultaneously the minimum acceptable system version (find_package) and the exact pinned version (FetchContent URL). Conflating "floor" with "pin" in one variable is a smell — and if the real intent was to consume system Boost on Ubuntu, the floor should be lowered (e.g. 1.74/1.83) rather than left at 1.87, which guarantees the source-build fallback everywhere.
4. The array{object{...}} fix has no inline rationale. The reason it can't be the simpler brace form lives only in ABeltramo's commit message. Without a one-line comment explaining the 1.91 brace-elision change, someone will "simplify" it back and silently break newer Boost. Same applies to the deliberate value_from(...) calls.

ABeltramo:
1. I've removed `EXACT` **AND** I've dropped the fetch for boost.json which caught errors in CI because we've effectively upgraded that package to `1.87`
2. True, but if we ever update the runners that would become important again
3. The real intent was to set it as the floor and allow new versions to be used, not to consume the older Boost.
4. Partially false: `value_from(...)` would make CI fail as proved in the 2 commits before. The `array{object{...}}` will fail a unit test when run with a newer Boost version (it failed on my machine).

All in all I don't see anything wrong, apart from the LLM assumptions 😅

src/moonlight-server/sessions/common.hpp
```diff
@@ -1,6 +1,71 @@
...
+#include <thread>
+
+#include <helpers/logger.hpp>
 
 namespace wolf::core::sessions {
 constexpr std::string_view VIRTUAL_SINK_PREFIX = "virtual_sink_";
+constexpr std::chrono::milliseconds DEFAULT_WAYLAND_SOCKET_WAIT_TIMEOUT = std::chrono::seconds(5);
+
+inline std::chrono::milliseconds get_wayland_socket_wait_timeout() {
+  if (const char *timeout_ms = std::getenv("WOLF_WAYLAND_SOCKET_WAIT_TIMEOUT_MS")) {
```
ABeltramo: Just a reminder to change this to `utils::get_env` and add it to the docs and Dockerfile

ABeltramo (changes requested):

ABeltramo (approved):

Issue #225, opened 2025-07-08
Unable to use second GPU for encoding (Intel + AMD)


HeiyuEP:
There are Internal data stream error when moonlight connect to steam.
graphics card:
```
0000:03:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Navi 21 GL-XL [Radeon PRO W6800]
```

docker-compose config:
```
version: "3"
services:
  wolf:
    image: ghcr.io/games-on-whales/wolf:stable
    environment:
      - XDG_RUNTIME_DIR=/tmp/sockets
      - HOST_APPS_STATE_FOLDER=/etc/wolf
      - WOLF_RENDER_NODE=/dev/dri/renderD129
      - WOLF_LOG_LEVEL=DEBUG
    volumes:
      - /etc/wolf/:/etc/wolf
      - /tmp/sockets:/tmp/sockets:rw
      - /var/run/docker.sock:/var/run/docker.sock:rw
      - /dev/:/dev/:rw
      - /run/udev:/run/udev:rw
    device_cgroup_rules:
      - 'c 13:* rmw'
    devices:
      - /dev/dri
      - /dev/uinput
      - /dev/uhid
    network_mode: host
    restart: unless-stopped
```

wolf debug log:
```
wolf_1  | [2025-07-08 10:16:10] 
wolf_1  | [2025-07-08 10:16:10] [ /etc/cont-init.d/10-setup_user.sh: executing... ]
wolf_1  | [2025-07-08 10:16:10] **** Configure default user ****
wolf_1  | [2025-07-08 10:16:10] Container running as root. Nothing to do.
wolf_1  | [2025-07-08 10:16:10] DONE
wolf_1  | [2025-07-08 10:16:10] 
wolf_1  | [2025-07-08 10:16:10] [ /etc/cont-init.d/15-setup_devices.sh: executing... ]
wolf_1  | [2025-07-08 10:16:10] **** Configure devices ****
wolf_1  | [2025-07-08 10:16:10] Exec device groups
wolf_1  | [2025-07-08 10:16:11] Adding user 'root' to groups: gow-gid-107,root
wolf_1  | [2025-07-08 10:16:11] DONE
wolf_1  | [2025-07-08 10:16:11] 
wolf_1  | [2025-07-08 10:16:11] [ /etc/cont-init.d/30-nvidia.sh: executing... ]
wolf_1  | [2025-07-08 10:16:11] 
[...]

PR #419, opened 2026-06-01
10-bit SDR streaming + DMA-BUF zero-copy via gst-interpipe context sharing

JBailes:
## Summary

Started as the AMD DMA-BUF zero-copy work split out of #397, and grew into the proper fix Abe asked for there: stop hand-pinning DMA-BUF formats, fix negotiation at the source, and build 10-bit SDR streaming on top.

### 1. Fix negotiation in gst-interpipe (Abe's option 1)
The root cause of the original caps trouble is that `gst-interpipe` forwards buffers and events between decoupled pipelines but **not queries**, so VA/DMABuf elements on either side of the interpipe can't share a context. Fixed upstream in **games-on-whales/gst-interpipe#2** (forward `GST_QUERY_CONTEXT` across the boundary; tested, 15/15). Pinned in `docker/gstreamer.Dockerfile` until that merges.

### 2. Drop the manual DMA-BUF intersection
With negotiation working, the config-time `get_dma_caps` intersection is unnecessary. The producer now advertises a generic `video/x-raw(memory:DMABuf)` buffer and the compositor + `vapostproc` negotiate the concrete drm-format at runtime (verified on Intel VA with real `waylanddisplaysrc`: lands on a 10-bit format and propagates across the interpipe). Removes the now-dead `get_dma_caps` / `get_dma_caps_runtime` / `parse_dma_drm_formats` helpers.

### 3. 10-bit SDR streaming (HEVC/AV1 Main10)
10-bit encoding reduces compression banding even for 8-bit source. It's decoupled from HDR: the client's `dynamicRangeMode=1` only selects the 10-bit *decode* profile; HDR *display* is a separate control-stream message Wolf never sends, so the client decodes 10-bit while staying SDR (BT.709).

- **Advertise** `H265_MAIN10` / `AV1_MAIN10` in `ServerCodecModeSupport`, gated on a Main10 pipeline existing. Also fixes a real bug: the AV1 format constants were `0x10000`/`0x20000` but the protocol (Limelight.h) defines them as `0x1000`/`0x2000`.
- **Negotiate**: read `x-nv-video[0].dynamicRangeMode`, select a Main10 pipeline for HEVC/AV1.
- **Encode**: P010 `video_params` + `profile=main-10` pipelines for nvcodec / va / va-LP / qsv (HEVC and AV1). The compositor renders 10-bit; `vapostproc` converts to NV12 (8-bit sessions) or P010 (Main10).

## Verification
- gst-interpipe#2: built + 15/15 tests, plus an Intel-VA DMABuf-across-interpipe negotiation test.
- Real `waylanddisplaysrc` on Intel VA: generic DMABuf negotiates 10-bit end-to-end across the interpipe.
- Wolf CI: Docker build, Linux build + tests, clang-format all green.
- **Pending**: end-to-end stream to a Moonlight client (decode/visual check) across vendors.

## Notes for reviewers
- Depends on **games-on-whales/gst-interpipe#2** (Dockerfile pinned to that commit; unpin once merged).
- **Config migration**: bumps the config to v8. Existing installs (v<=7) are migrated automatically: the gstreamer section is regenerated from the new default (with the 10-bit pipelines) while hostname/uuid/paired_clients/profiles are preserved, and the old config is backed up as `config.toml.v7.old`. New `config.v8.toml` is the make_includable source.


ABeltramo:
Just carrying over my comment from https://github.com/games-on-whales/wolf/pull/397#issuecomment-4591299161 for context:

Now that I think about this, maybe this needs a full re-think..  

The way it works in `gst-wayland-comp` is that basically we publish the actual DMA formats **at runtime** based on the `render_node` param. This is supported in Gstreamer, and it's actually encouraged since you can't know the actual formats until you actually pick a GPU. Now what would happen when you have 2 `vaapi` capable GPUs? `vapostproc` most likely would return the formats of `renderD128` but that might not be the one that the user wants to pick and so it'll cause issues down the line later on.  

In an ideal world Gstreamer would do all the caps intersection at runtime without having us to manually fiddle with it, the problem is that `gstinterpipe` doesn't support that [gst-interpipe#111](https://github.com/RidgeRun/gst-interpipe/issues/111), [gst-interpipe#165](https://github.com/RidgeRun/gst-interpipe/pull/165). So maybe it's time that we try again with the power of LLMs to fix this properly once and for all?  

----

I'd suggest re-evaluating the root cause for this and ultimately remove the manual DMA formats. I think we can either:
 - Fix `gst-interpipe` so that it properly shares the VA context + queries
 - Follow the current CUDA approach and just deal with the VA context manually between interpipes.

JBailes:
@ABeltramo Ready for PR, and it now includes 10bit.

ABeltramo:
I've rebased on top of latest `:stable` and gave it a quick test on my AMD 9070 I think we are missing something:

```
[GSTREAMER] Starting video producer: waylanddisplaysrc name=wolf_wayland_source render_node=/dev/dri/by-path/pci-0000:2d:00.0-render ! video/x-raw(memory:DMABuf), width=1920, height=1080, framerate=60/1 ! 
interpipesink sync=true async=false name=5733117292241342039_video max-buffers=1
[...]
INFO gstwaylanddisplaysrc::waylandsrc::imp: Supported DMA formats: ["AB4H:0x0200000000082305", "AB4H:0x0200000000042305", "AB4H:0x0200000000082405", "AB4H:0x0200000000082205", "AB4H:0x0200000000082105", "AB4H:0x0200000000000305", "AB4H:0x0200000000000a04", "AB4H:0x0200000000000105", "AB4H", "XB4H:0x0200000000082305", "XB4H:0x0200000000042305", "XB4H:0x0200000000082405", "XB4H:0x0200000000082205", "XB4H:0x0200000000082105", "XB4H:0x0200000000000305", "XB4H:0x0200000000000a04", "XB4H:0x0200000000000105", "XB4H", "AR30:0x0200000000082305", "AR30:0x0200000000042305", "AR30:0x0200000000082405", "AR30:0x0200000000082205", "AR30:0x0200000000082105", "AR30:0x0200000000000305", "AR30:0x0200000000000a04", "AR30:0x0200000000000105", "AR30", "XR30:0x0200000000082305", "XR30:0x0200000000042305", "XR30:0x0200000000082405", "XR30:0x0200000000082205", "XR30:0x0200000000082105", "XR30:0x0200000000000305", "XR30:0x0200000000000a04", "XR30:0x0200000000000105", "XR30", "AB30:0x0200000000082305", "AB30:0x0200000000042305", "AB30:0x0200000000082405", "AB30:0x0200000000082205", "AB30:0x0200000000082105", "AB30:0x0200000000000305", "AB30:0x0200000000000a04", "AB30:0x0200000000000105", "AB30", "XB30:0x0200000000082305", "XB30:0x0200000000042305", "XB30:0x0200000000082405", "XB30:0x0200000000082205", "XB30:0x0200000000082105", "XB30:0x0200000000000305", "XB30:0x0200000000000a04", "XB30:0x0200000000000105", "XB30", "AR24:0x0200000000082305", "AR24:0x0200000000042305", "AR24:0x0200000000082405", "AR24:0x0200000000082205", "AR24:0x0200000000082105", "AR24:0x0200000000000305", "AR24:0x0200000000000a04", "AR24:0x0200000000000105", "AR24", "AB24:0x0200000000082305", "AB24:0x0200000000042305", "AB24:0x0200000000082405", "AB24:0x0200000000082205", "AB24:0x0200000000082105", "AB24:0x0200000000000305", "AB24:0x0200000000000a04", "AB24:0x0200000000000105", "AB24", "XR24:0x0200000000082305", "XR24:0x0200000000042305", "XR24:0x0200000000082405", "XR24:0x0200000000082205", "XR24:0x0200000000082105", "XR24:0x0200000000000305", "XR24:0x0200000000000a04", "XR24:0x0200000000000105", "XR24", "XB24:0x0200000000082305", "XB24:0x0200000000042305", "XB24:0x0200000000082405", "XB24:0x0200000000082205", "XB24:0x0200000000082105", "XB24:0x0200000000000305", "XB24:0x0200000000000a04", "XB24:0x0200000000000105", "XB24", "AR15:0x0200000000082305", "AR15:0x0200000000042305", "AR15:0x0200000000082405", "AR15:0x0200000000082205", "AR15:0x0200000000082105", "AR15:0x0200000000000305", "AR15:0x0200000000000a04", "AR15:0x0200000000000105", "AR15", "RG16:0x0200000000082305", "RG16:0x0200000000042305", "RG16:0x0200000000082405", "RG16:0x0200000000082205", "RG16:0x0200000000082105", "RG16:0x0200000000000305", "RG16:0x0200000000000a04", "RG16:0x0200000000000105", "RG16", "R8  :0x0200000000082305", "R8  :0x0200000000042305", "R8  :0x0200000000082405", "R8  :0x0200000000082205", "R8  :0x0200000000082105", "R8  :0x0200000000000305", "R8  :0x0200000000000a04", "R8  :0x0200000000000105", "R8  ", "R16 :0x0200000000082305", "R16 :0x0200000000042305", "R16 :0x0200000000082405", "R16 :0x0200000000082205", "R16 :0x0200000000082105", "R16 :0x0200000000000305", "R16 :0x0200000000000a04", "R16 :0x0200000000000105", "R16 ", "GR88:0x0200000000082305", "GR88:0x0200000000042305", "GR88:0x0200000000082405", "GR88:0x0200000000082205", "GR88:0x0200000000082105", "GR88:0x0200000000000305", "GR88:0x0200000000000a04", "GR88:0x0200000000000105", "GR88", "GR32:0x0200000000082305", "GR32:0x0200000000042305", "GR32:0x0200000000082405", "GR32:0x0200000000082205", "GR32:0x0200000000082105", "GR32:0x0200000000000305", "GR32:0x0200000000000a04", "GR32:0x0200000000000105", "GR32"]
INFO waylanddisplaycore::utils::allocator: Creating DMA buffer - DrmFourcc: DrmFourcc(AB4H), Modifier: Unrecognized(144115188076389125)
```

with the generic `DMABuffer` cap, upstream we just settle on the first of the list to satisfy the constraint. Later on, the downstream pipeline starts:
```
Starting video pipeline: 
interpipesrc name=interpipesrc_5733117292241342039_video listen-to=5733117292241342039_video is-live=true stream-sync=restart-ts max-bytes=0 max-buffers=1 leaky-type=downstream !
vapostproc add-borders=true !
video/x-raw(memory:VAMemory), format=NV12, width=1920, height=1080, pixel-aspect-ratio=1/1 !
vah265enc aud=false b-frames=0 ref-frames=1 num-slices=1 bitrate=15308 cpb-size=255 min-qp=20 key-int-max=1024 rate-control=cbr target-usage=6 !
h265parse !
video/x-h265, profile=main, stream-format=byte-stream !
queue ! rtpmoonlightpay_video name=moonlight_pay payload_size=1392 fec_percentage=20 min_required_fec_packets=2 !
appsink sync=false max-buffers=1 drop=true name=wolf_udp_sink
09:33:26.209665635 DEBUG | Setting up wolf_udp_sink
09:33:26.209827105 DEBUG | Received NEED_CONTEXT for type gst.va.display.handle
0:00:16.541660311      54800      55269 WARN                GST_CAPS gstpad.c:5932:pre_eventfunc_check:<vapostproc3:sink> caps video/x-raw(memory:DMABuf), format=(string)DMA_DRM, width=(int)1920, height=(int)1080, framerate=(fraction)60/1, drm-format=(string)AB4H:0x0200000000082305 not accepted
0:00:16.556853989      54800      55269 WARN                 basesrc gstbasesrc.c:4262:gst_base_src_push_segment:<interpipesrc_5733117292241342039_video> segment format mismatched, ignore
0:00:16.556894109      54800      55269 WARN                GST_CAPS gstpad.c:5932:pre_eventfunc_check:<vapostproc3:sink> caps video/x-raw(memory:DMABuf), format=(string)DMA_DRM, width=(int)1920, height=(int)1080, framerate=(fraction)60/1, drm-format=(string)AB4H:0x0200000000082305 not accepted
0:00:16.556915839      54800      55269 WARN                GST_CAPS gstpad.c:5932:pre_eventfunc_check:<vapostproc3:sink> caps video/x-raw(memory:DMABuf), format=(string)DMA_DRM, width=(int)1920, height=(int)1080, framerate=(fraction)60/1, drm-format=(string)AB4H:0x0200000000082305 not accepted
0:00:16.556933429      54800      55269 WARN                 basesrc gstbasesrc.c:3187:gst_base_src_loop:<interpipesrc_5733117292241342039_video> error: Internal data stream error.
0:00:16.556942529      54800      55269 WARN                 basesrc gstbasesrc.c:3187:gst_base_src_loop:<interpipesrc_5733117292241342039_video> error: streaming stopped, reason not-negotiated (-4)
0:00:16.556982059      54800      55269 WARN                GST_CAPS gstpad.c:5932:pre_eventfunc_check:<vapostproc3:sink> caps video/x-raw(memory:DMABuf), format=(string)DMA_DRM, width=(int)1920, height=(int)1080, framerate=(fraction)60/1, drm-format=(string)AB4H:0x0200000000082305 not accepted
0:00:16.557005789      54800      55269 WARN                GST_CAPS gstpad.c:5932:pre_eventfunc_check:<vapostproc3:sink> caps video/x-raw(memory:DMABuf), format=(string)DMA_DRM, width=(int)1920, height=(int)1080, framerate=(fraction)60/1, drm-format=(string)AB4H:0x0200000000082305 not accepted
09:33:26.229087295 ERROR | [GSTREAMER] Pipeline error: Internal data stream error.
```

Ideally we should move `vapostproc` in the upstream pipeline so that `waylanddisplaysrc` and `vapostproc` agree on a format that they both can work with to produce 
```
video/x-raw(memory:VAMemory), format=NV12
```
so that downstream we just plug `vah265enc`.  

Now, this assumes that the VA context can travel with the latest gst-interpipe version but from a quick test that pipeline is failing for me. I'll dig deeper later when I have a bit of free time..

JBailes:
Damnit. I think I know the problem already, but I need to get home to fix it.

I'll have a fix in this afternoon. Not hard, but having to custom compile interpipe and run off that resulted in some missed stuff.

JBailes:
Okay, I already have a fix for this, but I don't like it. The fix I like is a much larger effort, and I am going with that.

ABeltramo:
On a tangent note: I've been testing the default pipeline on Nvidia with the new gst-interpipe and dropping `gst-video-context` and everything seems to run on my end. I'd very much like to drop all that extra stuff if we can confirm that this isn't required anymore.

JBailes:
> On a tangent note: I've been testing the default pipeline on Nvidia with the new gst-interpipe and dropping `gst-video-context` and everything seems to run on my end. I'd very much like to drop all that extra stuff if we can confirm that this isn't required anymore.

You hit exactly what I think the bigger fix is. It's going to take me a bit more time, but we'll get there!

ABeltramo:
Nice, I've pushed that since I had it ready locally. Feel free to revert if you still need some of that code..

JBailes:
Dependency on https://github.com/games-on-whales/gst-wayland-display/pull/34 but I suspect you'll be happy with the changes Abe.

ABeltramo:
Quick note before I forget: dropping the manual CUDA context management from Wolf would break multi Nvidia GPUs setups because we aren't setting `cuda-device-id` and the pipeline will default to device `0`. We should do the matching `WOLF_RENDER_NODE` -> `cuda-device-id` and set `waylanddisplaysrc` accordingly

ABeltramo:
So, elaborating on the [comment above](https://github.com/games-on-whales/wolf/pull/419#issuecomment-4740864798), here's where my mind is at (sorry for the long post 😅)

## Support Multi-GPU setups

- On Nvidia we need to control the `cuda-device-id` so that everything works on multi-GPU setups, and it's still 0 copy
- On AMD/Intel it's the same. Wolf is actually kind of broken at the moment OOB, and I've been suggesting to [manually exclude the second render device](https://github.com/games-on-whales/wolf/issues/225#issuecomment-3585056921) to get around this issue.
  - we should probably explore picking the right `varenderD*enc` which should take care of this

This may live here like we were doing for CUDA only, it might be possible to just have it as parameters of the pipeline, or it might make sense to do it directly in `gst-wayland-comp`, not sure what's the cleanest approach, but we need to stick everything to one GPU and sort this out once and for all.

## Caps exchange between gst-interpipe

I think here the best option is to move `cudaconvertscale` and `vapostproc` in the upstream pipeline. This will allow for a clean format exchange and agreement based on the real HW (assuming we pick the right GPU for both of them).

Re https://github.com/games-on-whales/gst-wayland-display/pull/34#issuecomment-4734630822 

> Two things stop me from reaching for it though. It pins format and resolution at the producer, so every consumer gets the same output and a resolution change means renegotiating the shared producer instead of just one branch, which is the fan-out interpipe is there to give us.

The resolution doesn't have to be pinned, only the format needs to be negotiated. You've already demonstrated that the resolution can be changed in real-time https://github.com/games-on-whales/wolf/pull/396 that means creating a new buffer, so I'm not even sure that the format _must be the same_.

> And passing the VA surface across the boundary isn't free: it's bound to a GstVaDisplay that doesn't propagate across the two pipelines, and the surface pool can recycle out from under the consumer since interpipe lets the pipelines run in different states. A DMABuf fd is much easier to hand over than a context-bound surface.

Yeah, a DMABuf is definitely easier to pass around and the right format for this; not sure about the complexity about passing a VABuffer, I've never looked into this honestly. If this is a hard blocker I guess we would need to make sure that we select the "right" DMA Buffer before starting the pipeline though. 

We could control the caps from the outside like we were doing in Wolf before these changes (maybe with a smarter way of doing it that it's more solid) or **we could add a custom Gstreamer plugin that does `DMA -> DMA` conversion** with shaders like `cudaconvert` (and I guess `vapostproc` do).   


----

@JBailes let me know your thoughts!

PR #420, opened 2026-06-04
fix(audio): set the pulse on_ready promise at most once

JBailes:
The PA context state callback resolves on_ready for READY/TERMINATED/FAILED. A boost::promise can only be set once, so a connection that goes READY
  then later TERMINATED (PulseAudio server going away) calls set_value twice and throws promise_already_satisfied, which is unhandled and crashes Wolf. Guard
  on_ready with a one-shot flag, reset in disconnect() when the promise is recreated.


JBailes:
PR 420 is MINE!

ABeltramo (approved):

PR #421, opened 2026-06-07
fix(fake-udev): persist /run/udev/data so libudev can enumerate devices

JBailes:
## Problem
`fake-udev` only broadcasts the `NETLINK_KOBJECT_UEVENT`; it never writes the `/run/udev/data/c<maj>:<min>` database record. libudev consumers enumerate devices from that database, not from the netlink event. In session images that run no `udevd` (gamescope/Steam), a Wolf virtual gamepad is `mknod`'d and its evdev events flow, but SDL never sees `ID_INPUT_JOYSTICK` for it, so certain games ignore the controller. Keyboard/mouse are unaffected because they don't depend on that tag. The fake-udev docs already state it should "generate the appropriate DB entries in `/run/udev/data/`" — the code never did.

## Evidence
- `strace fake-udev -m <add event>`: `socket`/`bind`/`sendmsg` only; no `open`/`write` under `/run/udev`.
- With the record absent, SDL (`udev_enumerate_scan_devices`) skips the pad; writing the record by hand makes Steam detect it on its next scan.
- A Wolf build whose `fake-udev` writes this record runs controllers correctly in production; this change brings stock `fake-udev` to parity.

## Change
On `add`, parse the uevent payload and write the udev DB entry the way `systemd-udevd` does: `E:` device properties, `I:USEC_INITIALIZED`, `G:`/`Q:` tags, `V:1`, keyed on `MAJOR:MINOR`. On `remove`, unlink it. Written before the broadcast, so a consumer that looks the device up in response to the event finds a complete record. No-op when the payload carries no `MAJOR`/`MINOR`. No new dependencies — `<fstream>`/`<sys/stat.h>`/`<unistd.h>` are already included.

### DIFF
--- a/src/fake-udev/fake-udev-cli.cpp
+++ b/src/fake-udev/fake-udev-cli.cpp
@@ -36,9 +36,8 @@ static void persist_udev_db(const std::string &msg) {
       tags = v;
     else if (k == "USEC_INITIALIZED")
       usec = v;
-    else if (!k.empty() && k[0] != '.' && k != "DEVPATH" && k != "SUBSYSTEM" && k != "DEVNAME" &&
-             k != "DEVTYPE" && k != "SEQNUM" && k != "DRIVER" && k != "MODALIAS" && k != "SYNTH_UUID" &&
-             k != "DEVLINKS")
+    else if (!k.empty() && k[0] != '.' && k != "DEVPATH" && k != "SUBSYSTEM" && k != "DEVNAME" && k != "DEVTYPE" &&
+             k != "SEQNUM" && k != "DRIVER" && k != "MODALIAS" && k != "SYNTH_UUID" && k != "DEVLINKS")
       props.emplace_back(k, v);
   }
 
--- a/src/fake-udev/fake-udev-cli.cpp
+++ b/src/fake-udev/fake-udev-cli.cpp
@@ -5,6 +5,81 @@
 #include <unistd.h>
 
 constexpr int UDEV_EVENT_MODE = 2;
+
+/**
+ * Broadcasting the uevent is not enough on its own: libudev consumers (e.g. SDL)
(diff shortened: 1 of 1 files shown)


ABeltramo:
We already do that here: https://github.com/games-on-whales/wolf/blob/c316027dc1ce07f226d7ea57c702bad1908ab436/src/moonlight-server/runners/docker.cpp#L256-L259 

Besides, you can't do that directly in `fake-udev` because it depends on the device, see `get_udev_hw_db_entries()` for the DualSense for example https://github.com/games-on-whales/wolf/blob/c316027dc1ce07f226d7ea57c702bad1908ab436/src/core/src/platforms/linux/uinput/joypad.cpp#L149-L205

JBailes:
> We already do that here:
> 
> https://github.com/games-on-whales/wolf/blob/c316027dc1ce07f226d7ea57c702bad1908ab436/src/moonlight-server/runners/docker.cpp#L256-L259
> 
> Besides, you can't do that directly in `fake-udev` because it depends on the device, see `get_udev_hw_db_entries()` for the DualSense for example
> 
> https://github.com/games-on-whales/wolf/blob/c316027dc1ce07f226d7ea57c702bad1908ab436/src/core/src/platforms/linux/uinput/joypad.cpp#L149-L205

Hmmm.

The current code is not working, the change is working.

ABeltramo:
What's currently not working? What's the actual issue?

JBailes:
Ah HAH, found the root cause bug. Thanks @ABeltramo !

JBailes:
> What's currently not working? What's the actual issue?

Certain games, such as Bravely Default's new remaster, aren't actually detecting controllers.

It's not an issue in Wolf based on your response, it's an issue in environment.

PR #422, opened 2026-06-07
Run PulseAudio inside the Wolf container

JBailes:
Fixes #374.

The external WolfPulseAudio sidecar has caused nothing but trouble. Wolf sleeps for `WOLF_PULSE_CONTAINER_TIMEOUT_MS` and then connects once, so you either set it too low and get no audio, or set it high (I had to use 30s on Unraid) and then any session started before the deadline gets no audio.

This runs PulseAudio inside the Wolf container and waits for the socket before launching Wolf, so PA is already up when Wolf does its existing "is there a PA server available?" check. Per @ABeltramo's suggestion it's Dockerfile-only, no C++ changes, and the sidecar stays as a fallback. Set `PULSE_SERVER` yourself to point Wolf at an external/host PA instead.

## Changes
- `wolf.Dockerfile`: install `pulseaudio` + `pulseaudio-utils`
- `startup.sh`: start PA on `$XDG_RUNTIME_DIR/pulse-socket`, wait for the socket, export `PULSE_SERVER` as the bare path, then exec Wolf. Skipped if `PULSE_SERVER` is already set
- docs: document `PULSE_SERVER`, note `WOLF_PULSE_IMAGE` is fallback-only

## Testing
Built `FROM ghcr.io/games-on-whales/wolf:stable` with these changes and ran the real Wolf binary:
- Wolf connects to the embedded PA on boot (`[PULSE] Pulse connection ready` -> `[PULSE_ROUTER] Setup complete`), no sidecar started, no "Unable to connect"
- App containers get `PULSE_SERVER` as the bare socket path they expect (`pa_context_get_server` returns it verbatim, no `unix:` prefix, so the socket mount still lines up)
- With PA removed, Wolf still falls back to the sidecar container as before

PR #423, opened 2026-06-07
Embed Wolf Den into the Wolf container

JBailes:
Follows the same pattern as #422 (embedded PulseAudio): instead of running Wolf Den as a separate `wolf-den` container, run it as another supervised process inside the Wolf container.

## Changes
- `wolf.Dockerfile`: copy the published Wolf Den app (`/app`) and the .NET runtime (`/usr/share/dotnet`) from `ghcr.io/games-on-whales/wolf-den:stable` (overridable via the `WOLF_DEN_IMAGE` build arg), install the `.NET` native deps not already present (`libstdc++6`, `zlib1g`), and symlink `dotnet` + point `/app/wolf-den` at the persisted state folder.
- `supervisord.conf`: add a `[program:wolf-den]` block. It waits for Wolf's control socket, then runs `dotnet WolfLeash.dll` on `:8080`. `priority=30` so it starts after Wolf and stops first on shutdown. autostart gated on `WOLF_EMBED_DEN`.
- `startup.sh`: default `WOLF_SOCKET_PATH=/var/run/wolf/wolf.sock` (Wolf creates it; Wolf Den connects to it via a `unix://` URI), create the Wolf Den state dir under the persisted state folder, and set `WOLF_EMBED_DEN=true` by default.

State (SQLite DB + data-protection keys) lives in `/etc/wolf/wolf-den`, so it survives container replacement. Set `WOLF_EMBED_DEN=false` to disable the embedded UI (e.g. if you run Wolf Den elsewhere).

## Validation (on a real Proxmox CT)
Built the embedded stack and ran it as a PVE CT:
- supervisord brings up `pulseaudio`, `wolf`, and `wolf-den`, all `RUNNING`.
- Wolf Den waits for Wolf's socket, connects, and serves: `Now listening on http://[::]:8080`, `Application started`; an HTTP probe returns `200`.
- `docker stop` shuts everything down gracefully in order: `wolf-den` -> `wolf` -> `pulseaudio`, each `exit 0`, in ~4s.

### DIFF
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -44,4 +44,17 @@ else
     export WOLF_EMBED_PULSE=false
 fi
 
+# Embedded Wolf Den: the .NET management UI runs as another supervised process
+# inside this container instead of the legacy separate "wolf-den" container.
+# Wolf creates its control socket at WOLF_SOCKET_PATH; Wolf Den connects to that
+# same socket (it wants a unix:// URI). Its state (SQLite DB + data-protection
+# keys) lives under the persisted state folder so it survives container replacement.
+export WOLF_SOCKET_PATH=${WOLF_SOCKET_PATH:-/var/run/wolf/wolf.sock}
+mkdir -p "$(dirname "$WOLF_SOCKET_PATH")"
+export WOLF_DEN_SOCKET="unix://$WOLF_SOCKET_PATH"
+mkdir -p "$HOST_APPS_STATE_FOLDER/wolf-den"
+# Enable the embedded Wolf Den by default; set WOLF_EMBED_DEN=false to disable it
+# (e.g. if you run Wolf Den elsewhere). Gates its autostart in supervisord.conf.
+export WOLF_EMBED_DEN=${WOLF_EMBED_DEN:-true}
+
 exec supervisord -c /etc/supervisord.conf
--- a/docker/supervisord.conf
+++ b/docker/supervisord.conf
@@ -1,8 +1,8 @@
-; Supervises the processes that make up Wolf. Today that's PulseAudio + Wolf;
-; drop in another [program:...] block to add more (e.g. Wolf Den).
+; Supervises the processes that make up Wolf: PulseAudio, Wolf, and the Wolf Den
+; management UI. Drop in another [program:...] block to add more.
 ;
-; Env (PULSE_SERVER, WOLF_EMBED_PULSE, ...) is exported by startup.sh and
-; inherited here. Program logs are forwarded to the container's stdout/stderr.
+; Env (PULSE_SERVER, WOLF_EMBED_PULSE, WOLF_EMBED_DEN, ...) is exported by
+; startup.sh and inherited here. Program logs go to the container stdout/stderr.
 
 [supervisord]
 nodaemon=true
@@ -52,6 +52,25 @@ startretries=3
 stopsignal=INT
 stopwaitsecs=30
 priority=20
+stdout_logfile=/dev/fd/1
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/fd/2
+stderr_logfile_maxbytes=0
+
+; Wolf Den, the .NET management UI, embedded here instead of the legacy separate
+; "wolf-den" container. autostart is gated on WOLF_EMBED_DEN so it can be turned
+; off. It waits for Wolf's control socket before starting (the dependency, no
+; fixed timeout), and connects to that socket directly via WOLF_SOCKET_PATH
+; (a unix:// URI, set per-program so it doesn't clash with the bare path Wolf
+; uses to create the socket). Highest priority, so on shutdown supervisord stops
+; it first (UI down before Wolf and PulseAudio). Listens on :8080.
+[program:wolf-den]
+command=bash -c 'sock="${WOLF_SOCKET_PATH#unix://}"; echo "[wolf-den] waiting for Wolf socket $sock"; until [ -S "$sock" ]; do sleep 0.5; done; cd /app && exec dotnet /app/WolfLeash.dll'
+autostart=%(ENV_WOLF_EMBED_DEN)s
+autorestart=true
+startretries=3
+priority=30
+environment=ASPNETCORE_HTTP_PORTS="8080",WOLF_SOCKET_PATH="%(ENV_WOLF_DEN_SOCKET)s",DOTNET_RUNNING_IN_CONTAINER="true",DOTNET_CLI_HOME="/tmp",HOME="/tmp"
 stdout_logfile=/dev/fd/1
 stdout_logfile_maxbytes=0
 stderr_logfile=/dev/fd/2
--- a/docker/wolf.Dockerfile
+++ b/docker/wolf.Dockerfile
@@ -107,6 +107,23 @@ RUN apt-get update -y && \
     pulseaudio pulseaudio-utils supervisor \
     && rm -rf /var/lib/apt/lists/*
 
+# Embedded Wolf Den: the .NET management UI runs as another supervised process
+# inside this container (see supervisord.conf + startup.sh) instead of the legacy
+# separate "wolf-den" container. We bring in the published app and the .NET
+# runtime straight from the Wolf Den image; its state (SQLite DB + data-protection
+# keys) lives in /etc/wolf/wolf-den (the persisted Wolf state folder), so it
+# survives container replacement. libstdc++6/zlib1g are the .NET native deps not
+# already pulled in above (libicu76, libssl3, ca-certificates are).
+ARG WOLF_DEN_IMAGE=ghcr.io/games-on-whales/wolf-den:stable
+RUN apt-get update -y && \
+    apt-get install -y --no-install-recommends \
+    libstdc++6 zlib1g \
+    && rm -rf /var/lib/apt/lists/*
+COPY --from=${WOLF_DEN_IMAGE} /usr/share/dotnet /usr/share/dotnet
+COPY --from=${WOLF_DEN_IMAGE} /app /app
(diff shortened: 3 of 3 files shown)


ABeltramo:
I'd say this should be done the other way around: instead of adding WolfDen as a dependency of Wolf why don't we build a uber image in WolfDen which contains both? It should be fairly simple too:
```Dockerfile
FROM ghcr.io/games-on-whales/wolf:stable

# Change supervisord and add WolfDen..
```

CC @Skerga

Skerga:
As a alternate image or the default? What are the main benefits for that? just the fewer configuration that would be needed, and the smaller compose?

ABeltramo:
Yeah, I'd say as an alternative image. Basically just to make the life easier for users (or platforms like Unraid) that want a battery included image with fewer knobs to set.  
Just thinking out loudly here, I'm very open to suggestions and ideas!

JBailes:
I'm already distributing a solution for those platforms though, Abe. That wouldn't represent a value-add to me, only a specialized image for maintaining.

I'd push to include Wolf's Den in Wolf proper more than creating Yet Another Image To maintain. I think we need to be looking at how to reduce the amount of images we are maintaining, not increasing.

Skerga:
in theory not that hard to make that change. But I would still use the current start process where a socket proxy is used so that wolf-den does not need to be ran as root. The socket would also be guaranteed to be available when socat tries the connection eliminating the risk of a not working connection. the only problem I can see is the mixing of the logs. should wolf-den follow the log settings for wolf or should the container have two envs for setting log level? And should i change the log format to fit more in line with the current wolf logs just so its prefixed with wolf-den?

ABeltramo:
> I'd push to include Wolf's Den in Wolf proper more than creating Yet Another Image To maintain

I personally don't like a fat image that includes `.net` and WolfDen; there are cases where it doesn't even make sense to have that included (ex: Fenrir for k8s). But, we had quite a few requests for a simpler setup and some users feel that it's "too complex" to spin up a stack for this. Plus, exposing the Wolf socket is quite a big attack vector..  

> But I would still use the current start process where a socket proxy is used so that wolf-den does not need to be ran as root [...] the only problem I can see is the mixing of the logs. should wolf-den follow the log settings for wolf or should the container have two envs for setting log level?

We've introduced `supervisord`, we can now set which user runs wolf-den, the log level, env variables and so on all in a simple config file, see `supervisord.conf` in this PR for example.

JBailes:
> > I'd push to include Wolf's Den in Wolf proper more than creating Yet Another Image To maintain
> 
> I personally don't like a fat image that includes `.net` and WolfDen; there are cases where it doesn't even make sense to have that included (ex: Fenrir for k8s). But, we had quite a few requests for a simpler setup and some users feel that it's "too complex" to spin up a stack for this. Plus, exposing the Wolf socket is quite a big attack vector..

In that case, then I'd push to just have a standard Wolf's Den distribution we recommend everyone to use.

ABeltramo:
> In that case, then I'd push to just have a standard Wolf's Den distribution we recommend everyone to use.

Not sure I'm following, what do you mean with that? A single image where `wolf:stable` becomes Wolf + Wolf Den?

ABeltramo:
How about 3 different tags:
- `wolf:minimal` which just contains Wolf + PA
- `wolf:with-web-ui` which contains Wolf + PA + WolfDen
- `wolf:stable` points to `wolf:with-web-ui` 

We should be able to easily do this in a single Dockerfile (aka our `wolf.Dockerfile`) with [docker build --target](https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage) and this way we keep both options. Then update the docs to add info about this.  

Thinking about it, it wouldn't make sense to move building Wolf + Den to Wolf Den because Wolf changes more frequently and we should always point to the latest `wolf:stable` when pushing an image that includes the web UI

JBailes:
> > In that case, then I'd push to just have a standard Wolf's Den distribution we recommend everyone to use.
> 
> Not sure I'm following, what do you mean with that? A single image where `wolf:stable` becomes Wolf + Wolf Den?

Having to maintain two Wolf's Den images is silly for this case, as well as our other images. End users would just get a recommendation to install a standard image that packages this together, in this case, it would be a recommendation to install Wolf's Den with a standard installation of Wolf-fedora43

ABeltramo:
> Having to maintain two Wolf's Den images is silly for this case

You are pulling from `ghcr.io/games-on-whales/wolf-den:stable` though so that wouldn't change anything we still have 2 Wolf Den images. 

> End users would just get a recommendation to install a standard image that packages this together

Sure, and that's why I think `wolf:stable` should point to this new image. I'd just like to still have a `wolf:minimal` tag that we can use; I don't think it'll be a maintainance burden, it's quite literally the same Dockerfile just with a different ENV variable and a couple less `copy --from`.

Hukuma1:
I'm all for some of the consolidation, but someone needs to take a look at Wolf-Den currently taking up a bit of CPU just idling. I sometimes kill it the container in my compose when not needing it due to this currently.

https://github.com/games-on-whales/wolf-den/issues/5

ABeltramo:
That's a very good point @Hukuma1 that should definitely be sorted first; I'll look into it

JBailes:
> That's a very good point @Hukuma1 that should definitely be sorted first; I'll look into it

We did an in-depth dive on this in discord, the answer IIRC was that this was "Just C#". 3%-5% or less of one cpu core is not outside of the realms of .NET idling.

Hukuma1:
My Home Assistant container which runs my entire home with over 100 devices has a usage of about 5% and idles down to 1%. Perhaps we can look into some sleep functionality for Wolf-Den? How often do we need the webUI past some initial setup and later if we add something? Maybe it can be brought up/down on-demand just like the other services now like PulseAudio, etc?

JBailes:
C# web services is the wrong language/platform for what you are asking.

Hukuma1:
Gotcha. Does Wolf-Den need to be based on C#?

If the user is going to be given a choice to run Wolf and Wolf-Den separately and save CPU cycles, I'd imagine that's the optimal path for some. But you want to consolidate the services for ease of setup (which I'm all for btw). I just ask you also please consider efficiency into your plan.

JBailes:
> Gotcha. Does Wolf-Den need to be based on C#?
> 
> If the user is going to be given a choice to run Wolf and Wolf-Den separately and save CPU cycles, I'd imagine that's the optimal path for some. But you want to consolidate the services for ease of setup (which I'm all for btw). I just ask you also please consider efficiency into your plan.

What Abe has discussed would mean you would want wolf-minimal. I'm very much of the opinion we need to reduce images, the test matrix right now is pretty crazy and I'm missing quite a few things in my code because it's too much of a surface right now. At least we need to make concerted efforts to standardize on fedora-43 (Well, fedora-44 but I'll get that PR out) and drop Ubuntu.

Hukuma1:
I personally would want _everyone_ to benefit. I would like to use the combined images, but not at the cost of unnecessary cycles/heat on my server. We can't have both?

nhths:
> the test matrix right now is pretty crazy

In my opinion, we should move away from specialized containers (Pegasus, Lutris, RetroArch) and consolidate everything into a single "emulators" container driven by theming. (Maybe need one "emulators" with multiple auto start scripts, which can be opened in lobby as now?)

wolf.sock should expose a stable API contract, and Wolf Den's only requirement is to support that contract. Den was built after wolfmanager, which is now archived. There's no need to lock ourselves to one specific UI - any client that speaks the contract is fine (CLI, custom frontend, scripted automation, etc.). And if Den or Wolf  breaks, we should be able to roll back to a previous version until a fix is released.

ABeltramo:
> In my opinion, we should move away from specialized containers (Pegasus, Lutris, RetroArch) and consolidate everything into a single "emulators" container driven by theming.

Which we already do, all emulators images share the same base installation and then we just add the single front-end package on top.

> wolf.sock should expose a stable API contract, and Wolf Den's only requirement is to support that contract. 

Already is like that, the API is stable and documented.

> Den was built after wolf-ui, which is now abandoned. 

Wrong, nothing is abandoned, they are two different projects.

This whole comment is wrong on so many levels. Let's keep the discussion on point guys.

Skerga:
Also there a quite some changes planned for wolf-ui the problem is that I'm currently swamped with other things and wolf-den changes where easily implemented compared to the planned things for wolf-ui

The current system is build in a way that nothing is required, you can build your own ui for lobbies and coop and just swap out wolf-ui. Wolf-ui just like wolf-den does everything via the socket. And wolf-ui does what it was build for: co-op, and non device bound save locations

nhths:
> Wrong, nothing is abandoned, they are two different projects.

Sorry, I mixed up wolf-ui and Wolf Manager. Wolfmanager was archived.

JBailes:
I almost want to close this, but I am not sure we reached consensus.

Was the agreement that this goes to @Skerga to implement? Or are there other action items?

ABeltramo:
I personally would like to have two images: minimal (which is what currently :stable is) and a bigger image that also embeds WolfDen so that users can choose which one to install based on their preference. I don't think this is a mantainance burden, it should really be a few lines in the Dockerfile.

Now I'd say the one with WolfDen should be the new `stable` but if it's using CPU when idle then we should address that first. If it's impossible to get C# to be fully idle (I doubt it) then that's a problem and we should consider something else for WolfDen.

JBailes:
In that case, I'm going to propose that I build out the minimal image, using Fedora-44 to base the future stable image on, and leave the wolf-den future :stable image to @Skerga as it sounds like you also want the future :stable to be gated on Wolf-Den's cpu utilization being reduced.

ABeltramo:
Yep, I think that's the right choice

JBailes:
Closing based on discussion

PR #426, opened 2026-06-12
Fix OpenAPI schema for lobby events

Baz00k:
While working with the Wolf API, I noticed that the OpenAPI schema does not match the actual emitted events. This PR fixes that.

### DIFF
--- a/docs/modules/dev/partials/spec.json
+++ b/docs/modules/dev/partials/spec.json
@@ -1847,7 +1847,7 @@
             "type": "string"
           },
           "session_id": {
-            "type": "integer"
+            "type": "string"
           },
           "encrypt_audio": {
             "type": "boolean"
@@ -1914,7 +1914,7 @@
             "type": "string"
           },
           "moonlight_session_id": {
-            "type": "integer"
+            "type": "string"
           },
           "pin": {
             "anyOf": [
@@ -1942,7 +1942,7 @@
             "type": "string"
           },
           "moonlight_session_id": {
-            "type": "integer"
+            "type": "string"
           }
         },
         "required": [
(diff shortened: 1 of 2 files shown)


src/moonlight-server/api/api.hpp
```diff
@@ -151,6 +151,17 @@ struct LobbyCreateResponse {
...
+  std::string lobby_id;
+  std::string moonlight_session_id;
+  std::optional<std::vector<short>> pin = std::nullopt;
+};
+
+struct LeaveLobbyRequest {
+  std::string lobby_id;
+  std::string moonlight_session_id;
+};
+
```
ABeltramo: Duplicate of  https://github.com/games-on-whales/wolf/blob/ea4a75d59deb9171dde3d21d6beb2785cb695854/src/moonlight-server/events/events.hpp#L180-L193 why have you added those?

ABeltramo (changes requested):
Not sure what exactly was missing from the `spec.json`, it looks like you've just replaced two structs with other two identical structs but the result should be exactly the same unless I'm missing something. Can you elaborate?

Baz00k:
Sure! Moonlight session id is now reported as string, not integer and matches the real returned type

Skerga:
The openapi spec does have some problems, i was planning on looking into them but never found the time
this fix would make [d088303](https://github.com/games-on-whales/GamesOnWhales.Wolf.Net.Bindings/commit/d088303ba8fed5dd8f9c8475293fea71448bb1af) obsolete

ABeltramo:
Thanks, I see the mismatch now.  

The underlying issue is that `size_t` can't be serialized as a number in JSON because it's `unsigned` and sending it as is will mess up most decoders. We should fix the code that generates json schema to avoid turning `size_t` into `"type": "integer"` instead of patching this with additional structs.

Baz00k:
@ABeltramo Something like this?

ABeltramo (approved):
LGTM, thanks for fixing this!

PR #427, opened 2026-06-13
Add support for explicit sync

ABeltramo:
Builds on top of https://github.com/games-on-whales/gst-wayland-display/pull/31 to add support for Wayland explicit sync by implementing `wp_linux_drm_syncobj_manager_v1`

TODO:

- [x] Test on different GPUs


ABeltramo:
I've tested this out both on AMD and Nvidia with a fairly high resolution/framerate `3440x1440@120`, disabled vsync in Moonlight and it _feels_ much better compared to `wolf:stable`. I don't have any hard data or recording, but I couldn't see any tearing and, most importantly, everything seems to work the same.

Would you be able to give it a go on your side @JBailes ?

JBailes:
👍  LGTM, especially since I wrote this and tested it prior ;)

PR #428, opened 2026-06-13
ci: add Fedora PR build images

JBailes:
Split out of #396 (party-mode split rendering), which had bundled unrelated CI infrastructure.

## What changed
- Fedora-based build images: `gpu-drivers`, `gstreamer`, and `wolf`
- a `pr-images` workflow that builds and tags per-PR images
- `docker-build` now also runs on pushes to `master`/`stable`

## Why
This is CI/build-image infrastructure unrelated to the party-mode feature; keeping it in its own PR makes both changes reviewable in isolation.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -2,6 +2,9 @@ name: Docker build
 
 on:
   push:
+    branches:
+      - master
+      - stable
     paths-ignore:
       - "docs/**"
       - ".github/**"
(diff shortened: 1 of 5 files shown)


JBailes:
@ABeltramo Split this out from party mode. It should be a pretty easy merge.

ABeltramo:
I think for Wolf we should just switch to Fedora, why would we keep the Ubuntu version?   

There's no risk of breaking any user image (or specific setup) here; as long as Wolf works it should work for everyone. I think this would also be a very good way to test out the Fedora images because everyone will pull `:stable` and test it out.

JBailes:
Superseded by #431

PR #429, opened 2026-06-13
feat: advertise comprehensive display modes

JBailes:
Split out of #396 (party-mode split rendering), which had bundled this unrelated catalog change.

## What changed
`DISPLAY_CONFIGURATIONS` now advertises a comprehensive mode catalog — 16:9 / 16:10 / 3:2, 21:9 and 32:9 ultrawide, and legacy 4:3 — generated from a resolution / refresh-rate table instead of a hand-maintained list.

## Why
Strict clients (e.g. Nintendo Switch) refuse to connect unless their native mode is advertised, so the catalog errs on the inclusive side. This is independent of party mode and reviewable on its own.

### DIFF
--- a/src/moonlight-server/state/data-structures.hpp
+++ b/src/moonlight-server/state/data-structures.hpp
@@ -22,6 +22,7 @@
 #include <optional>
 #include <state/serialised_config.hpp>
 #include <utility>
+#include <vector>
 
 namespace state {
 using namespace std::chrono_literals;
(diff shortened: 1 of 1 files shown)


JBailes:
@ABeltramo This needs to be merged before #396 , but I figured I'd split it out to make your job easier.

ABeltramo (approved):

PR #433, opened 2026-06-15
Update wayland client test

ABeltramo:
Brings back the Wayland client test and enables it in CI


ABeltramo:
This should be good to go, we are now down to 3 minutes with a warm cache.

JBailes:
@ABeltramo took a pass through this, few things worth flagging.

The big one: `TestWaylandDisplay`'s ctor blocks on `socket_future.get()`, but the error handler only quits the loop, it never sets the promise. So if the plugin fails to load (bad GST_PLUGIN_PATH, render-node=software unsupported, missing dep) the test doesn't fail, it just hangs until the 30min timeout. Kind of defeats the point of having it in CI. Should set the promise from `message::error`, or wait_for with a timeout.

CI versions look bumped a touch optimistically. `upload-artifact@v7` but `download-artifact@v8`, those usually move as a matched pair so worth confirming they're actually compatible. Same question for cache@v5 / checkout@v6 and the ubuntu-26.04 + arm runner labels being GA and not preview. If any of those are wrong the whole matrix hard fails before it does anything useful.

docker/CMakeLists.txt: swapping `target_link_libraries_system` for plain `target_link_libraries` loses the SYSTEM marking on the CURL/Boost.json includes. Anything building -Werror will start surfacing their header warnings as errors, worth a look on the clang job. The Boost::json PUBLIC promotion is right though, and the header-only to compiled switch is clean.

Rest is nits: TEST_RUST_WAYLAND is a misnomer now (no rust linking left, it's the gst plugin path), and the 65536 in the motion test would read better as a named constant.

ABeltramo:
> if the plugin fails to load (bad GST_PLUGIN_PATH, render-node=software unsupported, missing dep) the test doesn't fail

The pipeline is built before that, if the plugin is not available, or it fails to instantiate it throws and early stops the test. I agree on the timeout though, if for any reason there's no message sent this will hang; added it.

> CI versions look bumped a touch optimistically. upload-artifact@v7 but download-artifact@v8, those usually move as a matched pair so worth confirming they're actually compatible. Same question for cache@v5 / checkout@v6 and the ubuntu-26.04 + arm runner labels being GA and not preview. If any of those are wrong the whole matrix hard fails before it does anything useful.

Yeah, I've upgraded to latest on everything. It seems to be working though so not sure why we should bump back to a previous version. `26.04` is actually needed for the Wayland test because `24.04` bundles a too old version.

> docker/CMakeLists.txt: swapping target_link_libraries_system for plain target_link_libraries loses the SYSTEM marking on the CURL/Boost.json includes. Anything building -Werror will start surfacing their header warnings as errors, worth a look on the clang job. The Boost::json PUBLIC promotion is right though, and the header-only to compiled switch is clean.

True but I'd be surprised if Curl or Boost generates any warning, and if they do, we should probably address them rather than suppress. Plus we don't build with `-Werror` (and we probably should)..

JBailes (approved):

PR #439, opened 2026-06-21
fix(nvidia): zero-copy via dmabuftocuda

JBailes:
Nvidia zero-copy was broken. The producer pushed `memory:CUDAMemory` over the interpipe; CUDAMemory can't cross it (bound to a CUDA context the per-client consumer do't share), and `cudaupload` has no `memory:DMABuf` sink caps, so it can't import the dmabuf either.

VAAPI works because it carries a DMABuf over the interpipe. This makes Nvidia do the same.

**Fix (2 files):**
- producer caps: `memory:CUDAMemory` → `memory:DMABuf, drm-format=NV12`
- nvcodec `video_params_zero_copy`: `cudaupload` → `queue ! dmabuftocuda`

`dmabuftocuda` imports the dmabuf into CUDA in the consumer (EGLImage, no sysmem copy), where the encoder's CUDA context lives.

**Needs:** gst-wayland-display built with the cuda feature (provides `dmabuftocuda`). The Dockerfile pin predates it bump the pin.

**Tested** on an RTX 5080: the chain `waylanddisplaysrc → DMABuf(NV12) → interpipe → queue → dmabuftocuda → cudaconvertscale → CUDAMemory → nvh265enc` encodes H.265. Daemon boots logging `Using zero copy pipeline on Nvidia`. Not run: a live Moonlight session.

### DIFF
--- a/src/moonlight-server/state/configTOML.cpp
+++ b/src/moonlight-server/state/configTOML.cpp
@@ -301,7 +301,12 @@ Config load_or_default(const std::string &source,
   if (use_zero_copy) {
     switch (video_encoder) {
     case NVIDIA: {
-      default_base_video.producer_buffer_caps = "video/x-raw(memory:CUDAMemory)";
+      // Carry an NV12 DMABuf over the interpipe and import it into CUDA in the
+      // consumer (dmabuftocuda, see nvcodec video_params_zero_copy). A CUDAMemory
+      // buffer can't cross the interpipe -- it's tied to a CUDA context/stream the
+      // per-client encoder pipeline doesn't share -- so the producer must hand off
+      // the context-free dmabuf, exactly like the VAAPI path below.
+      default_base_video.producer_buffer_caps = "video/x-raw(memory:DMABuf), drm-format=NV12";
       break;
     }
     case VAAPI:
--- a/src/moonlight-server/state/default/config.v7.toml
+++ b/src/moonlight-server/state/default/config.v7.toml
@@ -339,8 +339,14 @@ video/x-raw(memory:CUDAMemory), width={width}, height={height}, \
 chroma-site={color_range}, format=NV12, colorimetry={color_space}, pixel-aspect-ratio=1/1\
 """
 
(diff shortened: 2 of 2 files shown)


JBailes:
Note that this finding is current, but is dependent on PR 37 from gst-wayland-display to merge.

ABeltramo:
> Nvidia zero-copy was broken

Not on my machine :sweat_smile: 

> The producer pushed memory:CUDAMemory over the interpipe; CUDAMemory can't cross it (bound to a CUDA context the per-client consumer do't share)

Wrong, we manually create the context in Wolf and pass it to both pipelines see `gst-video-context` and `need_context_for_device()`

JBailes:
> > Nvidia zero-copy was broken
> 
> Not on my machine 😅
> 
> > The producer pushed memory:CUDAMemory over the interpipe; CUDAMemory can't cross it (bound to a CUDA context the per-client consumer do't share)
> 
> Wrong, we manually create the context in Wolf and pass it to both pipelines see `gst-video-context` and `need_context_for_device()`

PR 37 Abe from gst-wayland-display, and pah, I just re-read the PR body. Was not sufficiently descriptive of PR 37 being involved. :/

ABeltramo:
Let's re-assess the changes needed in Wolf when https://github.com/games-on-whales/gst-wayland-display/pull/37 is finished and merged. The description is wrong, the comment in the code repeats the same hallucination. We don't actually _need_ to even touch the CUDA pipeline, let's start by fixing what's currently broken first :wink:

ABeltramo:
Funnily, with the changes that you've added in https://github.com/games-on-whales/gst-interpipe/pull/2 the CUDA context actually gets correctly shared between the two sides on the interpipe.

JBailes:
> Funnily, with the changes that you've added in [games-on-whales/gst-interpipe#2](https://github.com/games-on-whales/gst-interpipe/pull/2) the CUDA context actually gets correctly shared between the two sides on the interpipe.

Thanks to this comment, I found the root cause and fixed it. My harness had an old version of interpipe that it was building against, which meant...all the work here was worthless. 😢 

At least the harness is fixed now to make sure it always has the latest code pulled!

PR #442, opened 2026-06-22
feat: better control over motion joypads

kmreisi:
## Problem

Wolf trusts the controller type the Moonlight client reports. For phone clients that's often wrong: `moonlight-common-c` (used by all Moonlight clients) advertises `GYRO`/`ACCELEROMETER` capability whenever the phone has sensors enabled, but Moonlight Android specifically downgrades the **reported type** to `UNKNOWN` whenever the phone has sensors and the underlying pad isn't already a DualSense:

```java
// moonlight-android/app/src/main/java/com/limelight/binding/input/ControllerHandler.java#L3148
if (type != MoonBridge.LI_CTYPE_PS && sensorManager != null) {
    // Override the detected controller type if we're emulating motion sensors
    reportedType = MoonBridge.LI_CTYPE_UNKNOWN;
}
```

Wolf's default for UNKNOWN is the Xbox virtual pad. The motion-request blocks (`if (capabilities & GYRO|ACCELEROMETER && final_type == PS)`) only fire for PS pads — so the gyro/accel data the client is ready to send is silently dropped.

## Fix

Two related changes, bundled:

### 1. Auto-detect (always on, no config)

When the client reports `controller_type=UNKNOWN` AND advertises `GYRO` or `ACCELEROMETER`, Wolf creates a PS5/DualSense pad instead of falling through to Xbox. The motion blocks then fire and the gyro data is forwarded end-to-end. Stock Moonlight Android with any non-PS pad on any phone with sensors enabled now gets motion routed out of the box.

The promotion is unconditional because Moonlight Android itself shows an in-game toast ("Gamepad type may be changed due to motion sensor emulation") when the user enables motion — the client-side already signals the type change is expected.

### 2. Per-client `motion_controller_override` setting

A new optional `motion_controller_override: ControllerType` field on `ClientSettings` (default `AUTO`). Sibling of `controllers_override`: same shape of "operator wants type X" but gated on the client actually having motion to forward. When `controllers_override[slot]` is unset, the client advertises `GYRO`/`ACCELEROMETER`, AND `motion_controller_override != AUTO`, Wolf forces that pad type instead of the auto-detected one.

```
PATCH /api/v1/clients/<id>/settings  { "motion_controller_override": "PS" }
```

Same plumbing as `controllers_override` — added to `PartialClientSettings`, threaded through both the lobby-create and client-settings-update endpoints.

Useful for "every motion-capable connection from this client should land on device X" preferences. Forward-compatible with PR #385's Switch Pro work — once those motion-gates extend beyond `PS`, `motion_controller_override = "NINTENDO"` will route motion to a Switch Pro pad.

## Precedence

```
controllers_override[slot]   (per-slot)    ← highest, wins everything
        ↓ unset
motion_controller_override   (per-client)  ← forces type if motion advertised AND != AUTO
        ↓ AUTO
auto-detection                              ← UNKNOWN+motion → PS, else honour client's reported type
```

## Compatibility

Existing clients with `motion_controller_override` unset keep their behaviour for explicit XBOX/PS/NINTENDO. The only behaviour change is for UNKNOWN-with-motion clients, which previously got Xbox+no-motion and now get DualSense+motion — universally an improvement for the affected users (everyone on Moonlight Android with a non-DS pad + sensors enabled).

## Documented

In `docs/modules/user/pages/configuration.adoc` under "Override the default joypad mapping" — two new sub-sections covering the auto-detection and `motion_controller_override`.

## Verified

End-to-end on a Pixel 7 running stock Moonlight Android with an Xbox pad attached: client reports UNKNOWN with `GYRO|ACCELEROMETER`, Wolf creates a DualSense, and gyro/accel events arrive at the virtual pad — no config needed. With `motion_controller_override="XBOX"` set on the same client, Wolf creates an Xbox pad instead.

### DIFF
--- a/docs/modules/user/pages/configuration.adoc
+++ b/docs/modules/user/pages/configuration.adoc
@@ -325,6 +325,37 @@ The available joypad types are:
 * `nintendo`
 * `ps`
 
+==== Auto-detection: UNKNOWN clients with motion get a DualSense
+
+Moonlight Android downgrades the reported controller type to `UNKNOWN` whenever the phone has sensors and the underlying pad isn't already a DualSense (see https://github.com/moonlight-stream/moonlight-android/blob/master/app/src/main/java/com/limelight/binding/input/ControllerHandler.java[`ControllerHandler.java`]) — the capability bits still advertise `GYRO` and `ACCELEROMETER`. Wolf's auto-detection treats this as the client signalling "I have motion data to give" and creates a PlayStation DualSense pad so the motion-request blocks actually fire and the gyro/accel data gets forwarded. Without this the type would fall through to Xbox and the motion would be silently dropped.
+
+Auto-detection only kicks in when there's no per-slot `controllers_override` and no per-client `motion_controller_override` (below). Both of those are explicit operator choices and win.
+
+==== Per-client `motion_controller_override`
+
+Sibling of `controllers_override`: same shape of "operator wants type X" but gated on the client actually having motion to forward. When `controllers_override[slot]` is unset, the client advertises `GYRO` / `ACCELEROMETER`, AND `motion_controller_override` is set to anything other than `auto`, Wolf creates that pad type instead of the auto-detected one. Useful to point motion-capable clients at a non-default device — for example forcing a Switch Pro pad (once the corresponding motion routing in `create_new_joypad` lands).
+
+[source,toml]
+....
+[[paired_clients]]
+# client_cert = ...
+[paired_clients.settings]
+motion_controller_override = "PS" # motion-capable clients get a DualSense
+....
+
+Available values are the same as `controllers_override` (`auto` (default) / `xbox` / `nintendo` / `ps`); `auto` defers to the auto-detection above.
+
+Precedence: per-slot `controllers_override` > per-client `motion_controller_override` > auto-detection.
+
+It is also exposed over the API:
+
+[source,bash]
+....
+curl --unix-socket /var/run/wolf/wolf.sock \
+     -X PATCH http://localhost/api/v1/clients/<client_id>/settings \
+     -d '{"settings": {"motion_controller_override": "PS"}}'
+....
+
 === Input mouse overrides
 
 You can increase and decrease the mouse acceleration and scroll speed for each paired client, ex:
--- a/src/moonlight-server/api/api.hpp
+++ b/src/moonlight-server/api/api.hpp
@@ -58,6 +58,7 @@ struct PartialClientSettings {
   std::optional<float> mouse_acceleration;
   std::optional<float> v_scroll_acceleration;
   std::optional<float> h_scroll_acceleration;
+  std::optional<wolf::config::ControllerType> motion_controller_override;
 };
 
 struct UpdateClientSettingsRequest {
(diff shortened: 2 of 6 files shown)


ABeltramo (changes requested):
I like the direction where this is going, but I would take it one step further..  

If there's no manual override in the config (`final_type == AUTO`) and the client says that it supports `GYRO/ACCELEROMETER` I think we should have a better option than the current `bool motion_as_ds`: we should make that basically an enum that will inform Wolf which device should be picked. This would allow us to support different joypad types like the work that you've been doing in https://github.com/games-on-whales/inputtino/pull/37 

Other than that, the only other change is that we need to add this to the docs, possibly in the joypad section [here](https://games-on-whales.github.io/wolf/stable/user/configuration.html#_override_the_default_joypad_mapping) or in a separate section close to that one.

Btw, on my Android Moonlight client I can't get it to report `UNKNOWN`; it's always asking for an `XBOX` pad. Unchecking the `Xbox 360/One USB gamepad driver` option in the setting doesn't change that. Not sure how to replicate the `UNKNOWN` scenario but if you agree on the changes it doesn't really matter.

kmreisi:
Hi @ABeltramo,

I changed the behaviour according to your request (I hope I understood correctly ;-) ).

The option in Android to enable motion is "Emulate gamepad motion sensor support" (only available if the device supports it). When enabled, [ControllerHandler.java#L3148](https://github.com/moonlight-stream/moonlight-android/blob/master/app/src/main/java/com/limelight/binding/input/ControllerHandler.java#L3148) kicks in:

```java
byte reportedType;
if (type != MoonBridge.LI_CTYPE_PS && sensorManager != null) {
    // Override the detected controller type if we're emulating motion sensors on an Xbox controller
    Toast.makeText(activityContext, activityContext.getResources().getText(R.string.toast_controller_type_changed), Toast.LENGTH_LONG).show();
    reportedType = MoonBridge.LI_CTYPE_UNKNOWN;

    // Remember that we should enable the clickpad emulation combo (Select+LB) for this device
    needsClickpadEmulation = true;
}
```

With the latest changes:

- **Client does not send motion controls and is undefined** ⇒ XBOX (unchanged)
- **User enables motion controls, no override configured in Wolf** ⇒ PS
- **User enables motion controls, override configured in Wolf** ⇒ override

This kind of matches the notification seen in the Moonlight client: "Gamepad type may be changed due to motion sensor emulation".

I will work on the Switch Pro controller integration when I'm back home — here I have just Xbox controllers + Android phones. Together with the other NAT PR (#441) we did a 4-user Wolf LAN party from a host in Germany to an Airbnb in France :-D

Thanks,
Karim

kmreisi:
A couple of design questions I forgot to ask in the previous comment — happy to adjust either if you have a different preference:

**Naming.** I called the new field `motion_controller_override` to mirror the existing `controllers_override` (same shape of "operator wants type X", just gated on the client having motion). Open to a different name if `motion_controller_override` reads wrong to you — `motion_controller_type`, `motion_pad_override`, etc. are all easy renames.

**Relationship to `controllers_override`.** I kept `controllers_override` (per-slot, unconditional force) as the higher-priority knob, with `motion_controller_override` (per-client, motion-gated) below it. Precedence: per-slot `controllers_override` > per-client `motion_controller_override` > auto-detection. Does that match how you'd expect them to compose, or did you envision the new field as a slot-position vector too (`vector<ControllerType>` like `controllers_override`)? I went with a single value per client because the motion case is dominated by "phone with N emulated pads, all same caps" where per-slot granularity wouldn't add anything, but happy to make it a vector for symmetry if you'd prefer.

ABeltramo (approved):
Using the official Moonlight app from the Google Play Store (`v12.1`, last updated in 2024 👀) it seems that I only get `CONTROLLER_MULTI_PACKET` which doesn't support the joypad and capabilities of `CONTROLLER_ARRIVAL_PACKET`. Tried messing with the options in the config page but didn't really change much..

Anyway, I think these changes are good I like that users can now control the `motion_controller_override` and that we provide a better default if the client asks for `UNKNOWN`.

PR #446, opened 2026-06-24
Add a generic per-app properties map

kmreisi:
**Draft — a generic per-app `properties` map so wolf can carry app-specific settings without per-app schema changes.**

## What
Apps gain an optional `properties: map<string, string>` that wolf itself does **not** interpret — the app image / runner does (e.g. `steam_library_dirs`, emulator firmware paths). It flows end-to-end:

`BaseApp` (TOML config) → `events::App` (runtime) → `Reflector<App>` (socket API `/api/v1/apps`), and round-trips back to TOML on save.

## Why
External tooling (custom launchers / web UIs over the socket API) and app-specific runners need somewhere to attach settings meaningful only to a particular app, without growing wolf's core schema for each one. A generic string map keeps wolf agnostic while the app side interprets its own keys.

```toml
[[profiles.apps]]
title = "Steam"

[profiles.apps.properties]
steam_library_dirs = "/mnt/ssd/steam:/mnt/nas/steam"
```

## Notes
- Optional + defaults to empty → no behaviour change for existing configs.
- An empty map round-trips to *absent* (no empty TOML table is written back).
- First of a small series of hooks (this, plus client-settings policy fields) that let an external launcher/UI drive wolf entirely over the socket API.

### DIFF
--- a/src/moonlight-server/events/events.hpp
+++ b/src/moonlight-server/events/events.hpp
@@ -4,6 +4,7 @@
 #define BOOST_THREAD_PROVIDES_FUTURE
 #include <boost/thread.hpp>
 #include <boost/thread/future.hpp>
+#include <map>
 #include <core/audio.hpp>
 #include <core/input.hpp>
 #include <core/virtual-display.hpp>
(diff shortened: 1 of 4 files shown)


Skerga:
Isn't this already possible by setting env vars for each app?

kmreisi:
> Isn't this already possible by setting env vars for each app?

Yes, you are right, thanks for pointing out :smile: 

I'm trying to upstream / propose my custom wolf setup (rust based launcher + web +  custom docker images) there i used originally simpified config files for my apps, and there it was somehow needed...

But you are right the proper way is to simply use env variables, this can be closed :-).

PR #447, opened 2026-06-24
Per-client profile access in ClientSettings

kmreisi:
Adds per-client access control to `ClientSettings` — the existing home for client-specific config (`run_uid`, `controllers_override`, `motion_controller_override`, …).

## What
- `allowed_profiles: vector<string>` — profile IDs a client may see and launch. Empty (default) = unrestricted (all profiles).
- `show_coop_games: bool` (default `true`) — whether a client may see and join other clients' co-op (multi-user) lobbies.

Both are settable over the socket API through `PartialClientSettings` (the update-client-settings endpoint) and flow through create-lobby, exactly like the existing per-client fields. They show up in the clients response too.

## Why
On a shared / multi-tenant host, different clients should see different sets of profiles, and co-op visibility is naturally per-client. `ClientSettings` is already where per-client policy lives, so this slots in with no new mechanism. Defaults preserve today's behaviour (all profiles visible, co-op visible).

This lets an external UI that drives Wolf over the socket read a client's `allowed_profiles` / `show_coop_games` and filter what it shows — without Wolf needing to know anything about that UI.

Draft pending CI.

### DIFF
--- a/src/moonlight-server/api/api.hpp
+++ b/src/moonlight-server/api/api.hpp
@@ -59,6 +59,8 @@ struct PartialClientSettings {
   std::optional<float> v_scroll_acceleration;
   std::optional<float> h_scroll_acceleration;
   std::optional<wolf::config::ControllerType> motion_controller_override;
+  std::optional<std::vector<std::string>> allowed_profiles;
+  std::optional<bool> show_coop_games;
 };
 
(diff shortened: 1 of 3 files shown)


kmreisi:
Closing for now, will create a proper PR when ready, thanks :-)

PR #452, opened 2026-07-02
fix(lobby): deliver device unplug events to the lobby container

kmreisi:
## Problem

Hot-plugging a controller (disconnect then reconnect) while a Moonlight session is inside a lobby app (e.g. Steam) doesn't reach the app — the reconnected pad never shows up. A plain Wolf-UI session is fine; it only breaks once the session has joined a lobby.

## Cause

The lobby's `UnplugDeviceEvent` handler re-fires the event to the lobby's runner as a **bare** `events::UnplugDeviceEvent`:

```cpp
app_state->event_bus->fire_event(
    events::UnplugDeviceEvent{.session_id = lobby->id, ...});
```

Every handler registers for `immer::box<events::UnplugDeviceEvent>`, and `dp::event_bus` keys handlers by `typeid(EventType)`. So the unboxed event matches no handler and is silently dropped — the lobby container never receives the remove.

The stale device node then makes the next replug fail at `mknod ... File exists`, and since the plug command is `mknod ... && fake-udev`, the `&&`-chained ADD uevent never fires, so the app never sees the reconnect.

The sibling paths (plug relay, join, leave) all box the event; this one didn't — it has been this way since lobbies were introduced.

## Fix

Box the relayed event, matching the other paths. Includes a regression test that drives the relay through `setup_lobbies_handlers` (no compositor required).

### DIFF
--- a/src/moonlight-server/sessions/lobbies.cpp
+++ b/src/moonlight-server/sessions/lobbies.cpp
@@ -313,10 +313,10 @@ setup_lobbies_handlers(const immer::box<state::AppState> &app_state,
         immer::vector<events::Lobby> lobbies = app_state->lobbies->load();
         if (auto lobby = state::get_lobby_by_connected_session(lobbies, unplug_device_event->session_id)) {
           logs::log(logs::debug, "[LOBBY] Unplug device for session {}", unplug_device_event->session_id);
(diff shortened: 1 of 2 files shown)


ABeltramo (approved):
Thanks, LGTM

Issue #451, opened 2026-06-30
DualSense input leaks into host


lyubomirv:
Host system:
OS: Arch
DE: Gnome, on Wayland
GPU: Rx 7900 XT
Monitors: 2 monitors

Client:
Chromecast with Google TV
DualSense controller

I'm running Wolf and have two apps installed - Steam and Heroic. Initially I've noticed that when I'm playing a game the controller's touchpad (which works fine in the game) also controls the mouse pointer on the host. When testing further using a browser-based gamepad tester, I actually see that I get all input from the controller on my host's desktop.

I initially saw this in the Heroic app and thought that it might be something related to it but then tested Steam as well and even just the Wolf UI - it's all the same. I can see the DualSense being detected as a Wolf DualSense by the testing web page on the host as soon as I open the Wolf UI app in Moonlight, and disconnected as soon as I close it.

I thought input should be isolated, so am I doing something wrong?

The only settings that I've changed in the TOML config file are:
- The name of the host
- Force the controller to be a PS one for the first client
- Map the host dirs where my Steam/Heroic games are installed for those app containers

JBailes:
See https://github.com/games-on-whales/unraid-plugin/issues/57  for other reports of this.

lyubomirv:
I'm not sure this is the same. I'm not using unraid, and I'm not experiencing leaking from apps into wolf ui and starting other apps. Of course, it might still be the same issue, though, just saying.

ABeltramo:
Have you added the udev rules as described in the [quickstart guide](https://games-on-whales.github.io/wolf/stable/user/quickstart.html#_virtual_devices_support)? With those applied, the joypads will only be available to `root` and it'll block access to any other user effectively achieving isolation. Unless you log in your system as `root` that is..

lyubomirv:
Yes, I have. And I am not logging into my desktop as root.

However, what I noticed back when I added those rules, is that it was a kind of an odd location for my system. The directory `/etc/udev/rules.d/` was completely empty and these rules file was the first thing I added. I remembered that Sunshine also has some similar rules file and found that it is located in `/usr/lib/udev/rules.d/` - where I found a huge amount of other rules files. So this seemed like the proper location.
I tried moving the wolf rules there but it was exactly the same.

[...]

PR #453, opened 2026-07-04
Change the name of the PS5 controller in the rules file to DualSense

lyubomirv:
Change the name of the PS5 controller in the rules file to DualSense in order to match what the device is registered as.

The current rules file doesn't catch the DualSense correctly which leads to it being detected by the host and its input is leaked.

Found in #451


lyubomirv:
Seems like this is superseeded by #455. Closing.

PR #454, opened 2026-07-06
fix(input): stop Wolf virtual controllers leaking into the host desktop

### LINKED ISSUE #451: DualSense input leaks into host

JBailes:
## Problem

Wolf's virtual gamepads are created via `uinput`/`uhid` inside the Wolf container but appear as **real devices in the host kernel**. The udev rules that are supposed to keep them off the host's active seat don't do their job, so `systemd-logind` grants the logged-in desktop user a `uaccess` ACL and the streamed controller **also drives the host desktop** (e.g. the DualSense touchpad moves the host mouse pointer).

Fixes #451. Also matches the reports in games-on-whales/unraid-plugin#57.

## Root cause

Three compounding defects in `85-wolf.rules` (and the copy in the quickstart docs):

1. **Stale device names.** The rules match `Wolf PS5 (virtual) pad` and `Wolf gamepad (virtual) motion sensors`, but the pad is created as `Wolf DualSense (virtual) pad` (`src/moonlight-server/control/input_handler.cpp:126`), and the kernel's `hid-playstation` driver names its children `… pad Touchpad` and `… pad Motion Sensors`. None of those match, so the rules never fire for the DualSense.
2. **No rule for the DualSense touchpad** child device — so it kept moving the host cursor even after a user corrected the pad name.
3. **The actual leak vector.** The joypad rules set only `MODE`/`GROUP`, with no seat reassignment and no `uaccess` removal. A `uaccess` ACL overrides the group permissions, so a pad leaked to the host **even when its name matched** — which is why the leak reproduced across Xbox, DualSense and the Wolf UI.

## Fix

- Match every Wolf virtual device by its **`Wolf ` name prefix (glob)** instead of hard-coded names — covers the gamepad, the DualSense `Touchpad` + `Motion Sensors` child devices, and the `hidraw` node, and won't go stale on a future rename.
- Match both the device's own name (`ATTR`, for the uinput pads) and any ancestor's name (`ATTRS`, for the uhid DualSense children + hidraw).
- Park the devices on the phantom **`seat9`** and **strip the `uaccess` ACL** so no logged-in host user can read them.
- Update the quickstart docs to match, and add a troubleshooting entry.

## New diagnostic — `scripts/wolf-input-diag.sh`

Host-side, non-destructive tool to validate controller passthrough and isolation (it never grabs a device, so it can't steal input from a running game):

- `host` — audits every `Wolf …` device on the host and flags any reachable by a non-root user or not covered by the rules (the leak in this issue).
- `discover` / `static` — lists per-container input devices and fails if a node is shared across containers.
- `monitor` / `live` — live checks that input reaches the container it should and no other.

It doubles as the acceptance test for the rules change.

## Testing

- Device names verified against source (`input_handler.cpp`) and the pinned `inputtino` dependency; rules and script are syntax-checked.
- ⚠️ **Needs confirmation on hardware with a physical DualSense.** I don't have a way to assert the exact `logind`/`udev` behaviour (whether `seat9` + `TAG-="uaccess"` fully strips the ACL on the touchpad child device) without a real device. Reproduce and confirm with the bundled diagnostic:
  ```bash
  sudo ./scripts/wolf-input-diag.sh host   # BEFORE: LEAK lines for the pad + touchpad
  sudo cp 85-wolf.rules /etc/udev/rules.d/ && sudo udevadm control --reload-rules && sudo udevadm trigger
  # reconnect the controller, then:
  sudo ./scripts/wolf-input-diag.sh host   # AFTER: all root-only, rule-covered
  ```

### DIFF
--- a/85-wolf.rules
+++ b/85-wolf.rules
@@ -1,14 +1,27 @@
-# Allows Wolf to access /dev/uinput
-KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput"
+# Allows Wolf to access /dev/uinput (needed to create the virtual gamepad/keyboard/mouse)
+KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput", TAG+="uaccess"
 
-# Allows Wolf to access /dev/uhid
-KERNEL=="uhid", TAG+="uaccess"
+# Allows Wolf to access /dev/uhid (needed for DualSense emulation)
+KERNEL=="uhid", MODE="0660", GROUP="input", TAG+="uaccess"
 
-# Move virtual keyboard and mouse into a different seat
+# Move the virtual keyboard and mouse onto a phantom seat (seat9) so the host's
+# logged-in desktop user cannot read them.
 SUBSYSTEMS=="input", ATTRS{id/vendor}=="ab00", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9"
 
-# Joypads
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf X-Box One (virtual) pad", MODE="0660", GROUP="input"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf PS5 (virtual) pad", MODE="0660", GROUP="input"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf gamepad (virtual) motion sensors", MODE="0660", GROUP="input"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf Nintendo (virtual) pad", MODE="0660", GROUP="input"
+# Wolf virtual controllers: the gamepads (Xbox / DualSense / Nintendo), the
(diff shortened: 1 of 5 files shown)


JBailes:
Superseded by #455 — moved onto the main repo with cleaned-up commit history. Closing this one.

PR #455, opened 2026-07-06
fix(input): stop Wolf virtual controllers leaking into the host desktop

### LINKED ISSUE #451: DualSense input leaks into host

JBailes:
## Problem

Wolf's virtual gamepads are created via `uinput`/`uhid` inside the Wolf container but appear as **real devices in the host kernel**. The udev rules that are supposed to keep them off the host's active seat don't do their job, so `systemd-logind` grants the logged-in desktop user a `uaccess` ACL and the streamed controller **also drives the host desktop** (e.g. the DualSense touchpad moves the host mouse pointer).

Fixes #451. Also matches the reports in games-on-whales/unraid-plugin#57.

## Root cause

Three compounding defects in `85-wolf.rules` (and the copy in the quickstart docs):

1. **Stale device names.** The rules match `Wolf PS5 (virtual) pad` and `Wolf gamepad (virtual) motion sensors`, but the pad is created as `Wolf DualSense (virtual) pad` (`src/moonlight-server/control/input_handler.cpp:126`), and the kernel's `hid-playstation` driver names its children `… pad Touchpad` and `… pad Motion Sensors`. None of those match, so the rules never fire for the DualSense.
2. **No rule for the DualSense touchpad** child device — so it kept moving the host cursor even after a user corrected the pad name.
3. **The actual leak vector.** The joypad rules set only `MODE`/`GROUP`, with no seat reassignment and no `uaccess` removal. A `uaccess` ACL overrides the group permissions, so a pad leaked to the host **even when its name matched** — which is why the leak reproduced across Xbox, DualSense and the Wolf UI.

## Fix

- Match every Wolf virtual device by the **`Wolf *virtual*` name glob** instead of hard-coded names — covers the gamepad, the DualSense `Touchpad` + `Motion Sensors` child devices, and the `hidraw` node, and won't go stale on a future rename.
- Match both the device's own name (`ATTR`, for the uinput pads) and any ancestor's name (`ATTRS`, for the uhid DualSense children + hidraw).
- Park the devices on the phantom **`seat9`** and **strip the `uaccess` ACL** so no logged-in host user can read them.
- Add a troubleshooting entry for the symptom.

## `85-wolf.rules` is now the single source of truth

Per review feedback, the rules were duplicated between `85-wolf.rules` and the quickstart docs. Rather than delete the file — downstream installers (e.g. the Unraid plugin, games-on-whales/unraid-plugin#57) want a canonical URL to pull from — the file stays and **the docs now fetch it instead of restating it**:

```bash
curl -fsSL https://raw.githubusercontent.com/games-on-whales/wolf/stable/85-wolf.rules \
  -o /etc/udev/rules.d/85-wolf.rules
```

This also fixes drift the two copies had already accumulated: the docs told users to install the rules as `85-wolf-virtual-inputs.rules`, while `scripts/wolf-input-diag.sh` checks `/etc/udev/rules.d/85-wolf.rules` — so the diagnostic would report "not rule-covered" on a correctly configured host.

Also dropped the stale claim that Wolf creates virtual keyboard/mouse devices — those events go to [gst-wayland-display](https://github.com/games-on-whales/gst-wayland-display), not `uinput`.

## New diagnostic — `scripts/wolf-input-diag.sh`

Host-side, non-destructive tool to validate controller passthrough and isolation (it never grabs a device, so it can't steal input from a running game):

- `host` — audits every `Wolf …` device on the host and flags any reachable by a non-root user or not covered by the rules (the leak in this issue).
- `discover` / `static` — lists per-container input devices and fails if a node is shared across containers.
- `monitor` / `live` — live checks that input reaches the container it should and no other.

It doubles as the acceptance test for the rules change.

## Testing

- Device names verified against source (`input_handler.cpp`) and the pinned `inputtino` dependency; rules and script are syntax-checked.
- ⚠️ **Needs confirmation on hardware with a physical DualSense.** The exact `logind`/`udev` behaviour (whether `seat9` + `TAG-="uaccess"` fully strips the ACL on the touchpad child device) should be confirmed on a real device. Reproduce and confirm with the bundled diagnostic:
  ```bash
  sudo ./scripts/wolf-input-diag.sh host   # BEFORE: LEAK lines for the pad + touchpad
  sudo cp 85-wolf.rules /etc/udev/rules.d/ && sudo udevadm control --reload-rules && sudo udevadm trigger
  # reconnect the controller, then:
  sudo ./scripts/wolf-input-diag.sh host   # AFTER: all root-only, rule-covered
  ```

### DIFF
--- a/85-wolf.rules
+++ b/85-wolf.rules
@@ -14,8 +14,8 @@ KERNEL=="uhid", MODE="0660", GROUP="input", TAG+="uaccess"
 # the host desktop. See https://github.com/games-on-whales/wolf/issues/451
 #
 # We match on the "Wolf *virtual*" name glob: every device Wolf creates is named
-# "Wolf ... (virtual) ..." (or "Wolf mouse virtual device"), so this covers all
-# current and future sub-devices without matching unrelated "Wolf" hardware.
+# "Wolf ... (virtual) ...", so this covers all current and future sub-devices
+# without matching unrelated "Wolf" hardware.
 # We then (a) park them on the phantom seat9 and (b) strip the uaccess ACL. Both
 # the device's own name (ATTR, for uinput pads) and any ancestor's name (ATTRS,
 # for the uhid DualSense child devices and hidraw node) are matched.
--- a/85-wolf.rules
+++ b/85-wolf.rules
@@ -1,12 +1,8 @@
-# Allows Wolf to access /dev/uinput (needed to create the virtual gamepad/keyboard/mouse)
+# Allows Wolf to access /dev/uinput (needed to create the virtual gamepad)
 KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput", TAG+="uaccess"
 
 # Allows Wolf to access /dev/uhid (needed for DualSense emulation)
 KERNEL=="uhid", MODE="0660", GROUP="input", TAG+="uaccess"
-
-# Move the virtual keyboard and mouse onto a phantom seat (seat9) so the host's
-# logged-in desktop user cannot read them.
-SUBSYSTEMS=="input", ATTRS{id/vendor}=="ab00", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9"
 
 # Wolf virtual controllers: the gamepads (Xbox / DualSense / Nintendo), the
 # DualSense's "Touchpad" and "Motion Sensors" child input devices, and the
@@ -17,11 +13,12 @@ SUBSYSTEMS=="input", ATTRS{id/vendor}=="ab00", MODE="0660", GROUP="input", ENV{I
 # desktop user an ACL (uaccess) on them, so the streamed controller also drives
 # the host desktop. See https://github.com/games-on-whales/wolf/issues/451
 #
-# We match on the "Wolf " name prefix (glob) so every current and future
-# sub-device is covered regardless of the exact product name, then (a) park them
-# on the phantom seat9 and (b) strip the uaccess ACL. Both the device's own name
-# (ATTR, for uinput pads) and any ancestor's name (ATTRS, for the uhid DualSense
-# child devices and hidraw node) are matched.
-SUBSYSTEM=="input",  ATTR{name}=="Wolf *",  MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf *", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
-KERNEL=="hidraw*",   ATTRS{name}=="Wolf *", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
+# We match on the "Wolf *virtual*" name glob: every device Wolf creates is named
+# "Wolf ... (virtual) ..." (or "Wolf mouse virtual device"), so this covers all
+# current and future sub-devices without matching unrelated "Wolf" hardware.
+# We then (a) park them on the phantom seat9 and (b) strip the uaccess ACL. Both
+# the device's own name (ATTR, for uinput pads) and any ancestor's name (ATTRS,
+# for the uhid DualSense child devices and hidraw node) are matched.
+SUBSYSTEM=="input",  ATTR{name}=="Wolf *virtual*",  MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Wolf *virtual*", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
+KERNEL=="hidraw*",   ATTRS{name}=="Wolf *virtual*", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
--- a/85-wolf.rules
+++ b/85-wolf.rules
@@ -1,14 +1,27 @@
-# Allows Wolf to access /dev/uinput
-KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput"
+# Allows Wolf to access /dev/uinput (needed to create the virtual gamepad/keyboard/mouse)
+KERNEL=="uinput", SUBSYSTEM=="misc", MODE="0660", GROUP="input", OPTIONS+="static_node=uinput", TAG+="uaccess"
 
-# Allows Wolf to access /dev/uhid
-KERNEL=="uhid", TAG+="uaccess"
+# Allows Wolf to access /dev/uhid (needed for DualSense emulation)
+KERNEL=="uhid", MODE="0660", GROUP="input", TAG+="uaccess"
 
-# Move virtual keyboard and mouse into a different seat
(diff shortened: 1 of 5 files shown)


docs/modules/user/pages/quickstart.adoc
```diff
@@ -733,18 +733,23 @@ crw------- 1 root root 10, 223 Jan 17 09:08 /dev/uinput
 .Create `udev` rules under `/etc/udev/rules.d/85-wolf-virtual-inputs.rules`
 [source,bash]
 ....
-# Allows Wolf to acces /dev/uinput (only needed for joypad support)
+# Allows Wolf to access /dev/uinput (needed to create the virtual gamepad/keyboard/mouse)
```
ABeltramo: We don't create keyboard and mouse devices via inputtino. We directly pass the events to [gst-wayland-display](https://github.com/games-on-whales/gst-wayland-display)

docs/modules/user/pages/quickstart.adoc
```diff
@@ -733,18 +733,23 @@ crw------- 1 root root 10, 223 Jan 17 09:08 /dev/uinput
...
+KERNEL=="uhid", MODE="0660", GROUP="input", TAG+="uaccess"
 
-# Joypads
-KERNEL=="hidraw*",   ATTRS{name}=="Wolf PS5 (virtual) pad", GROUP="root", MODE="0660", ENV{ID_SEAT}="seat9"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf X-Box One (virtual) pad", GROUP="root", MODE="0660", ENV{ID_SEAT}="seat9"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf PS5 (virtual) pad", GROUP="root", MODE="0660", ENV{ID_SEAT}="seat9"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf gamepad (virtual) motion sensors", GROUP="root", MODE="0660", ENV{ID_SEAT}="seat9"
-SUBSYSTEMS=="input", ATTRS{name}=="Wolf Nintendo (virtual) pad", GROUP="root", MODE="0660", ENV{ID_SEAT}="seat9"
+# Move the virtual keyboard and mouse onto a phantom seat (seat9)
+SUBSYSTEMS=="input", ATTRS{id/vendor}=="ab00", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9"
```
ABeltramo: Not sure what this covers, we don't create virtual keyboard and mouse devices.

docs/modules/user/pages/quickstart.adoc
```diff
@@ -733,18 +733,23 @@ crw------- 1 root root 10, 223 Jan 17 09:08 /dev/uinput
...
+
+# Wolf virtual controllers: the gamepads plus, for the DualSense, its "Touchpad"
+# and "Motion Sensors" child devices and its /dev/hidraw* node. Matching the
+# "Wolf " name prefix (glob) covers every current and future sub-device, then
+# parks them on seat9 and strips the uaccess ACL so the host desktop user can't
+# read them. See https://github.com/games-on-whales/wolf/issues/451
+SUBSYSTEM=="input",  ATTR{name}=="Wolf *",  MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
```
ABeltramo: `Wolf *` might be a bit too broad, any device that starts with Wolf will hit this rule..

85-wolf.rules
```diff
@@ -1,14 +1,27 @@
...
+# the host desktop. See https://github.com/games-on-whales/wolf/issues/451
+#
+# We match on the "Wolf " name prefix (glob) so every current and future
+# sub-device is covered regardless of the exact product name, then (a) park them
+# on the phantom seat9 and (b) strip the uaccess ACL. Both the device's own name
+# (ATTR, for uinput pads) and any ancestor's name (ATTRS, for the uhid DualSense
+# child devices and hidraw node) are matched.
+SUBSYSTEM=="input",  ATTR{name}=="Wolf *",  MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
+SUBSYSTEMS=="input", ATTRS{name}=="Wolf *", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
+KERNEL=="hidraw*",   ATTRS{name}=="Wolf *", MODE="0660", GROUP="input", ENV{ID_SEAT}="seat9", TAG-="uaccess"
```
ABeltramo: I never liked having the same info both here and in the docs, should we just get rid of this file? It's not actually used anywhere IIRC

ABeltramo (changes requested):
Just took a glance at the rules, I'll test out the scripts on my machine in a few..

85-wolf.rules
JesterEE: It may be beneficial to keep this file for other deployments/plugins to pull from.

For example, the current implementation in the Unraid plugin hard codes these rules in the installer script rather than referencing an expected configuration from wolf. Keeping this file updated within the wolf project and having installers pull this file during install could alleviate misconfigured application deployments. https://github.com/games-on-whales/unraid-plugin/issues/57#issuecomment-5004821790

JBailes:
@ABeltramo Should have fixed most of your feedback. I'm leaving 85 in there for now until I can come up with a better solution for all consuming downstream projects.

JBailes:
@ABeltramo , I should have addressed all your comments (As well as other related comments). Let me know what you are thinking here.

docs/modules/user/pages/quickstart.adoc
```diff
@@ -730,23 +730,22 @@ ls -la /dev/uinput
 crw------- 1 root root 10, 223 Jan 17 09:08 /dev/uinput
 ....
 
-.Create `udev` rules under `/etc/udev/rules.d/85-wolf-virtual-inputs.rules`
+The `udev` rules live in
+https://github.com/games-on-whales/wolf/blob/stable/85-wolf.rules[`85-wolf.rules`]
+in the Wolf repository; that file is the single source of truth, so install it
+directly rather than copying the rules by hand:
```
ABeltramo, line 736: In the new docs the inline copy-paste block is gone, a reader arriving fresh has never been told to copy anything by hand. 😅 

Maybe we can change it to something like 

> The `udev` rules live in https://github.com/games-on-whales/wolf/blob/stable/85-wolf.rules[`85-wolf.rules`] in the Wolf repository. You can install them directly with:

ABeltramo (approved):
I think this is good and I like how the docs reference the repo file instead of copying so that we have one source of truth. I can't test it on a real DualSense right now, but it looks good to me

PR #495, opened 2026-09-02
feat(gstreamer): patch nvh265enc to add intra-refresh support. Also updates gstreamer to 1.28.6 from August 5.

### LINKED ISSUE #494: No intra-refresh option on the NVENC pipelines — no equivalent to Sunshine's setting
Sunshine/Apollo expose an "intra refresh" encoder option that some clients (notably Moonlight on Xbox) rely on. I couldn't find a way to get the same thing out of Wolf on an NVIDIA host, none of the NVENC elements in the Wolf image seem to accept an intra-refresh property.

This is a hard requirement for the Xbox client, not having it sometimes corrupts the image with a color washing effect. intra-refresh should let the client self correct (currently only fixed by hard restart)

The Xbox client developer Andy has provided instructions on what needs to be done in this [discord](https://discord.com/channels/352065098472488960/856542329724993536/1544116451483459654) conversation.

Attaching his instructions here for reference as well:

yep, Wolf should enable intra-refresh in the encoder when the client requests `x-ss-video[0].intraRefresh=1`. You are just using ffmpeg right? 
For nvenc, ffmpeg just needs `-g 300 -intra-refresh 1 -single-slice-intra-refresh 1`. Just convert that to gstreamer I guess. For AMD, ffmpeg only supports it for h264 maybe you'll be lucky and it's already in GST.
You can also look for that and send a normal stream with regular keyframes  but that will result in a more bursty stream since those frames are very large.

andygrundman:
I took a guess at checking the SDP param and the TOML template but I didn't test it. 

It looks a bit weird that we have both` intra-refresh` and `single-slice-intra-refresh` but this is to match the options in ffmpeg. `gop-size=300` matches Sunshine but could be tuned based on how fast you want the recovery to happen.

I'll submit another PR for h264 and av1 if I get these accepted upstream.


docker/gstreamer.Dockerfile
RoeeKa: Needs to be removed

RoeeKa: Needs to be removed

RoeeKa (commented):
Thanks for this work Andy, I did manage to get this running with a locally built gstreamer. There are minor changes needed.

1. The build flags for gstreamer were [changed](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/9200) in 1.28, two of them had to be deleted (see comments)

2. Also please add 2 lines to `src/moonlight-server/streaming/streaming.cpp` in ‎`start_streaming_video` - This file isn't part of this PR so I couldn't leave an inline suggestion. 
The pipeline template is read from the config and substituted per stream at runtime, and `fmt::runtime` throws on a placeholder with no matching argument so these are needed.

```
fmt::arg("gop_size", video_session->gop_size),
fmt::arg("intra_refresh", video_session->intra_refresh),
```

I did verify that the properties reach the encoder so seems like things are wired correctly.

I am still not certain whether the issue itself on the client is fixed. I can say that I have seen the image corruption once at the start of my first stream but it may have been an old client state on my side. I did reboot and it hasn't happened since yet so it is reassuring.

PR #506, opened 2026-09-20
Apps added over the API never get a display

tonybart1337:
An app registered through `POST /api/v1/apps/add` never reaches a compositor.

`Reflector<events::App>` does not carry `video_producer_buffer_caps`, so an API-added app is constructed with an empty string, and `start_video_producer` then formats:

```
waylanddisplaysrc name=wolf_wayland_source render_node=/dev/dri/renderD128 ! , width=1280, height=720, framerate=60/1 !
interpipesink ...
```

`gst_parse_launch` rejects that (`[GSTREAMER] Pipeline parse error (recovered): syntax error`), the Wayland socket is never created, the app container's Xwayland exits with `could not connect to wayland server`, and Moonlight shows *No video received from host*.

Apps loaded from `config.toml` are unaffected: `configTOML.cpp` assigns `default_video_settings.producer_buffer_caps` to each of them. That is also why the same app streams fine right after a Wolf restart and breaks again the next time a client re-adds it over the API.

**Change:** `video_producer_buffer_caps` becomes part of the reflector, optional on the wire. An add that omits it inherits the caps from an app that already has them — the value follows from the encoder Wolf picked at startup and from zero copy, which no API client can know, and `endpoint_CreateSession` already inherits it from a sample app the same way. With no app to inherit from it falls back to `video/x-raw` rather than an empty string.

Reproduced and verified on Linux/NVIDIA (RTX 5090, nvcodec encoders) with a Moonlight client: before, every API-added app failed as above; after, the same app streams and its runner starts.


src/moonlight-server/events/reflectors.hpp
```diff
@@ -75,10 +83,17 @@ template <> struct Reflector<events::App> {
...
   }
 
+  static constexpr std::string_view DEFAULT_PRODUCER_BUFFER_CAPS = "video/x-raw";
+
   static events::App to(const ReflType &app, const std::shared_ptr<events::EventBusType> &ev_bus) {
     auto runner = Reflector<events::Runner>::to(app.runner, ev_bus);
+    auto buffer_caps = app.video_producer_buffer_caps.value_or(std::string{DEFAULT_PRODUCER_BUFFER_CAPS});
+    if (buffer_caps.empty()) {
+      buffer_caps = std::string{DEFAULT_PRODUCER_BUFFER_CAPS};
+    }
```
ABeltramo, line 93: We should never default to `video/x-raw`, I'd propose that `video_producer_buffer_caps` is not optional, users of the API can pick up the default from another populated app with an extra `GET` request.

ABeltramo (changes requested):
Thanks for contributing!  

I agree on the `Reflector` changes, just a couple more points that needs to be addressed before merging:
- `docs/modules/dev/partials/spec.json` needs to be updated manually (unfortunately)
- let's drop the default `video/x-raw`, see the comment below.

PR #507, opened 2026-09-20
Non-root app containers cannot connect to the Wayland socket

tonybart1337:
The compositor creates its socket `0755 root:root`. `connect(2)` on a unix socket requires **write** permission, so any app container that sets `User` (i.e. does not run as root) fails immediately:

```
could not connect to wayland server
(EE) Fatal server error:
(EE) Couldn't add screen
```

Wolf's own apps run as root, so this is invisible with the bundled app list; it shows up as soon as an external client registers an app that runs the game as the host user. The PulseAudio socket sitting next to it is already `srwxrwxrwx` for exactly this reason.

**Change:** widen the socket to `0777` right where `wait_for_wayland_socket` has just confirmed it exists, before the runner is started, and log a warning instead of failing if the permission change does not go through.

Verified on Linux/NVIDIA with a container running as `1000:1000`: before, Xwayland could not connect; after, it starts and the game runs. Same container as root worked both before and after, which is what pointed at the mode.


ABeltramo:
Thanks for taking the time to contribute!  

We solve this in GOW images by starting the container as `root`, fixing up the permissions for the unprivileged `retro` user for all required devices (controlled via `GOW_REQUIRED_DEVICES`) and then finally switching to the "less privileged" user.  
The change here would only affect the wayland socket and wouldn't fix for example access to `/dev/dri/renderX` (which in most OS installations is not accessible by default) or any input device in `/dev/input`. I'm not against an overhaul of the whole process, but this PR isn't the right way of doing it so I'm going to close it.

Issue #504, opened 2026-09-17
Wolf G_IS_OBJECT critical failure


r0b0tb4c0n:
when i make a connection with moonlight and then press the x in the top right corner, make a second connection pressing the x again, when i try to connect a third time i get the following:
```
wolf  | 17:40:20.988723740 DEBUG | Setting up wolf_udp_sink
wolf  | 0:00:53.535381756   165 0x7f375815ca80 WARN                basesink gstbasesink.c:1248:gst_base_sink_query_latency:<10907566951849915040_video> warning: Pipeline construction is invalid, please add queues.
wolf  | 0:00:53.535387316   165 0x7f375815ca80 WARN                basesink gstbasesink.c:1248:gst_base_sink_query_latency:<10907566951849915040_video> warning: Not enough buffering available for  the processing deadline of 0:00:00.020000000, add enough queues to buffer  0:00:00.020000000 additional data. Shortening processing latency to 0:00:00.000000000.
wolf  | 17:40:20.997562862 DEBUG | Context already set, passing it to the pipeline.
wolf  |
wolf  | (wolf:165): GLib-GObject-CRITICAL **: 18:40:20.997: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
wolf  |
wolf  | (wolf:165): GLib-GObject-CRITICAL **: 18:40:20.997: g_object_get: assertion 'G_IS_OBJECT (object)' failed
wolf  | 17:40:20.997623065 DEBUG | Wayland source message: WAYLAND_DISPLAY = wayland-2
wolf  | 17:40:20.997630619 INFO  | Wayland display ready, listening on: wayland-2
wolf  | 17:40:20.997745875 DEBUG | [STREAM_SESSION] Start runner
wolf  |
wolf  | (wolf:165): GStreamer-Cuda-CRITICAL **: 18:40:21.001: GstBufferPool* gst_cuda_buffer_pool_new(GstCudaContext*): assertion 'GST_IS_CUDA_CONTEXT (context)' failed
wolf  | 2026-09-17T17:40:21.001984Z  WARN gstwaylanddisplaysrc::waylandsrc::imp: Failed to create CUDA buffer pool: Failed to create CUDA buffer pool
wolf  | 17:40:21.002589767 DEBUG | /dev/dri/renderD129 vendor: NVIDIA Corporation
wolf  | 17:40:21.002606117 INFO  | Mounting nvidia driver nvidia-driver-vol:/usr/nvidia
[...]

Issue #501, opened 2026-09-14
Zero-copy NVIDIA pipeline crashes (GsCUDABuf/apply_video_info) on 2nd+ app launch — WOLF_USE_ZERO_COPY=FALSE works around it


jdonaldson:
## Summary

Under the zero-copy NVIDIA pipeline (default, `WOLF_USE_ZERO_COPY` unset), the first app launched in a Wolf process streams fine, but **every subsequent app launch in the same process crashes** with a panic in `GsCUDABuf`/`apply_video_info`, and the crash is severe enough that it sometimes segfaults the whole `wolf` process (exit status 11), auto-restarted by supervisord.

**Workaround confirmed**: setting `WOLF_USE_ZERO_COPY=FALSE` avoids the crash entirely — verified stable with two app containers (`Wolf-UI` and a custom `Steam` app) running concurrently, where the zero-copy path had crashed on the second launch every time it was tried.

Filing this because `AGENTS.md` documents that zero-copy is supposed to "auto-fall back to legacy when an encoder can't support it" — that fallback doesn't appear to be triggering here even though the zero-copy path is reliably broken in this environment, so I think the auto-fallback logic (rather than just the crash itself) is the actual bug.

## Environment

- GPU: RTX 4090
- Driver: 550.163.01 (proprietary, not open kernel modules)
- Host: Debian 13 (trixie), standard `linux-image-amd64` kernel (6.12.107+deb13-amd64) — **not** the Debian cloud kernel flavor, which lacks `uinput` and working `nvidia_drm` modeset entirely (separate issue, not filing that one, just noting for reproduction)
- Wolf image: `ghcr.io/games-on-whales/wolf:stable`
- Installation method: the documented "NVIDIA manual method" (`gow/nvidia-driver` volume + `NVIDIA_DRIVER_VOLUME_NAME`), not `nvidia-container-toolkit`/`runtime: nvidia` — switched to this specifically because the container-toolkit method hit the same `GsCUDABuf` crash on the very first launch (matches #379), and the manual method fixed *that* specific case
- Docker 29.8.0, `network_mode: host`, container run `privileged: true`

## Secondary finding, possibly related

[...]

PR #510, opened 2026-09-21
fix(rest): don't send a second response after a successful resume

### LINKED ISSUE #509: resume() sends a second HTTP response (400) after a successful resume
## Summary

`endpoints::https::resume()` writes **two** complete HTTP responses onto the same connection when a resume succeeds: the `200` + resume XML, immediately followed by a `400` from `server_error()`. The trailing `server_error()` call sits at function scope instead of inside the `else`.

## Code

[`src/moonlight-server/rest/endpoints.hpp#L442-L474`](https://github.com/games-on-whales/wolf/blob/2273b2e64868e6a144daa225a85e85b216d3af2c/src/moonlight-server/rest/endpoints.hpp#L442-L474) (stable @ `2273b2e`):

```cpp
  if (old_session) {
    ...
    auto xml = moonlight::launch_resume(rtsp_ip, std::to_string(get_port(state::RTSP_SETUP_PORT)));
    send_xml<SimpleWeb::HTTPS>(response, SimpleWeb::StatusCode::success_ok, xml);
  } else {
    logs::log(logs::warning, "[HTTPS] Received resume event from an unregistered session, ip: {}", client_ip);
  }

  server_error<SimpleWeb::HTTPS>(response);   // <-- also runs on the success path
}
```

`send_xml()` calls `response->write(...)` and sets `close_connection_after_response = true` (`rest/helpers.hpp#L42-L50`), and `server_error()` is itself a `send_xml()` with `client_error_bad_request`. So a successful resume emits `200 <resume xml>` followed by `400 <error xml>` on the same connection.

Three things suggest this is a leftover rather than intent:

- `launch()` and `cancel()` both end with a single `send_xml()` and no trailing `server_error()`.
- The other three `server_error<SimpleWeb::HTTPS>(response)` calls in this file (L349, L356, L419) are all inside conditionals; this is the only one at function scope.
- `git blame` puts the line in `cb0e168e` (2025-04-04) — the same commit that added the success-path `send_xml()` two lines above.

## Impact

I want to be precise about what I have and have not checked:

- **Verified by reading the source:** two responses are written on the success path.
- **Not verified at runtime:** how any particular Moonlight client reacts. It may well ignore the trailing bytes, given the connection closes right after.

Independent of client tolerance, it makes a successful resume indistinguishable from a failed one at the protocol level, which is awkward precisely when debugging the resume path.

## Suggested fix

Move the `server_error()` call inside the `else`. PR to follow.

## Context

Found while investigating re-entry into an already-running session (#501, #504). I am not claiming this causes the CUDA-context crashes in those issues — it is a separate defect noticed along the way.

## Environment

Read against `stable` @ `2273b2e`. Running Wolf built from that commit on Debian 13, RTX 4090, NVIDIA driver 550.163.01 (proprietary).

---
<sub>Analysis assisted by Claude Code; the code references and `git blame` above were checked against the repository before filing.</sub>

jdonaldson:
Fixes #509.

`resume()` ended with an unconditional `server_error<SimpleWeb::HTTPS>(response)` at function scope, so the success path wrote two complete HTTP responses onto the same connection: `200` + resume XML from `send_xml()`, then `400` + error XML from `server_error()`.

`send_xml()` calls `response->write(...)` and sets `close_connection_after_response = true`, and `server_error()` is itself a `send_xml()` with `client_error_bad_request` — so both were emitted back to back.

This moves the call into the `else`, which matches `launch()`, `cancel()`, and the other three `server_error()` call sites in this file (all already inside conditionals). `git blame` dates the line to `cb0e168e` (2025-04-04), the same commit that added the success-path `send_xml()` two lines above, so it looks like a leftover from that refactor.

**Verification**

- Compiles clean against `stable` @ `2273b2e` (built via `docker/wolf.Dockerfile --target wolf-builder`, 0 errors).
- Behaviour change verified by reading only. I have not instrumented a Moonlight client to confirm how it reacted to the trailing `400` — it may well have been ignoring it, since the connection closes right after. What the change guarantees is that a successful resume is now distinguishable from a failed one at the protocol level.
- No test added: the resume path needs a paired client and a live session, which the existing suite does not stand up. Happy to add one if you can point me at the right fixture.

Found while investigating re-entry into an already-running session (#501, #504). Not claiming it causes the CUDA-context crashes there — separate defect, noticed along the way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ABeltramo (approved):
Thanks for contributing, LGTM!

